PR #73, opened 2023-07-14
Enabling/disabling inputs

ashleybot:
Enhances validation and ui behavior (aka. enabling/disabling inputs, also called form logic). 

## Validation hierarchy

- Some validation, such as range validation should run no matter the form status
- However, *required* and *required if* properties should only fire validation if the form is attempting to be completed (aka. status = Complete)
- Only optional forms can be saved with a status of 'Not included' and in that case the reason code is required.
- Some forms have a different set of required fields for the initial visit vs follow-up visits. This validation is handled in the IsValid method rather than as an attribute.

## Forms with behaviors

These forms are implemented:

- [x] A1
- [x] A4
- [x] B1
- [x] B4
- [x] B5
- [x] B6
- [x] B7
- [x] B8
- [x] C2

These forms still need to be done:

- [ ] A2
- [ ] A3
- [ ] A5
- [ ] B9
- [ ] C1
- [ ] D1
- [ ] D2


smiththay (approved):
Looks great! Enabling/disabling inputs functions the way they should. Good work!

PR #74, opened 2023-07-17
Updates  LEARNED to have required if attribute

### LINKED ISSUE #70: A1 referral source validation
In the A1, `LEARNED` should be required if `REFERSC` is 1 or 2.

Image

If the user attempts to Complete then the form should display a validation error message near the LEARNED label.

smiththay:
- [x] LEARNED is required if REFERSC is 1 or 2
fixes #70


ashleybot (approved):
Looks good!

PR #77, opened 2023-07-27
Fixes bug with reason code mapping on not included forms

ashleybot:
- [x] Only optional forms should be allowed to have NotIncluded selected as a form status
- [x] If NotIncluded is selected then ReasonCode is required when saving the form
- [x] When a form that has NotIncluded as status is edited, the previously selected reason code should also be displayed in the drop down


smiththay (changes requested):
For me, it looks like if NotIncluded is selected then ReasonCode is **not** required when saving the form, except for A3 for some reason.

smiththay (approved):
Looks good!

PR #89, opened 2023-08-24
C2 t layout telephone visit kinds

### LINKED ISSUE #63: C2T layout for telephone visit kinds
**Follow-up telephone visits use a modified version of the C2 called the C2T**
Depending on the visit type, differing fields are required.

**Each visit type has it's own required set of forms**
The visit contract `UDS.Net.Services.DomainModels.Visit.BuildFormsContract()` defines which forms are required/optional for which visit types.

- [ ] IVP, required C2
- [ ] FVP, required C1 or C2
- [ ] TIP, required C2T
- [ ] TFP, optional C2T

**Additional context**
Depends on https://github.com/UK-SBCoA/uniform-data-set-dotnet-api/issues/2

smiththay:
- [x] C2T layout matches NACC form and has correct functionality and validation (except for RESPOTHX). RESPOTHX bug card needs to be deployed before this field will function correctly. 
Fixes #63


Oddvocado (changes requested):
Looks good! However, I seem to be running into a validation issue when attempting to save both the TFP and TIP forms as complete. Appears as if the MOCATRAI and MOCAREGI fields have validation that is being enforced on both forms.

Capture

Oddvocado (approved):
Looks good!

PR #90, opened 2023-08-28
Existing legacy id check

Oddvocado:


ashleybot (approved):
@Oddvocado Looks great!

I was able to:

- [ ] Correctly be prompted with an error message and link to existing profile when the legacy id already existed
- [ ] Modify the legacy id (to a new one) and create a new participation successfully.

Screen Shot 2023-08-30 at 8 40 26 AM

Before you merge this PR, be sure to update the dependency version numbers for the API nuget packages to 1.5.0 instead of the pre-release versions (this means the api package action will need to be finished first!)

PR #91, opened 2023-08-31
A2 form UI Behavior and Finalization

Oddvocado:


smiththay (changes requested):
@Oddvocado Looks good but there are a few things that need to be changed. ```INEDUC``` range should be 0-36 or 99 and ```INKNOWN``` range should be 0-120 or 999 according to NACC documentation [https://files.alz.washington.edu/documentation/uds3-ivp-ded.pdf](url). Also it seems the form can be completed without an input and without  validation to many of the fields except for ```INBIRMO``` and ```INBIRYR``` .

smiththay (approved):
Looks good!

PR #93, opened 2023-09-14
Saving form with disabled properties

Oddvocado:


smiththay (approved):
Looks like its functioning now!

PR #96, opened 2023-09-29
B9 Form Finalization

Oddvocado:


Oddvocado:
@ashleybot good catch! I've added the radio buttons and updated the questions of the follow up visits.

smiththay (approved):
Looks great! Functions correctly. The only thing I would add, maybe not all necessary but good for UX is to add the labels to direct user when parts of the form are skipped as it looks in the NACC documentation. For example: 
Screenshot 2023-10-09 at 10 23 28 PM
Screenshot 2023-10-09 at 10 23 45 PM


Again, I'm not sure if this is necessary, so up to you.

smiththay:
Good to merge!

PR #97, opened 2023-10-06
D1 Form Finalization

Oddvocado:
This PR aims to finalize the D1 UDS Form. Notable changes include

- [x] Adding client side validation
- [x] Adding server side validation
- [x] I added a "fancy checkboxes" stimulus controller. The intended function is to only allow users to select one checkbox for questions 5a-5e
- [x] Section 3 questions have now been updated to only allow the users to select 1 diagnosis as primary. If the participant has normal cognition, the primary, contributing and non-contributing fields will not store a value when the form is saved.
- [x] Section 3 questions have also been updated to require that users mark at least one diagnosis as present when participants do not have normal cognition.


smiththay (commented):
@Oddvocado Nice work on this complex form, especially the fancy checkboxes and section 3 validations. Only noticed one thing though. For questions  5a-5e , I can still click on the radio buttons even though I have not selected a checkbox. Validation works but it may be easier for the user if these radio buttons were disabled until the corresponding checkbox has been selected.
Screenshot 2023-10-31 at 4 52 35 PM

Oddvocado:
@smiththay Good catch! The radio buttons are now disabled when a checkbox is changed. Ready for re-review when you have a moment.

smiththay (commented):
@Oddvocado looks good! @ashleybot I am going to pass this review to you to do a final look through.

mlan225:
Looks good to me 👍

PR #98, opened 2023-10-12
D2 form

### LINKED ISSUE #85: D2 Form UI Behavior and Finalization

smiththay:
- [x] Form has correct format
- [x] Form has correct UI behavior for enabling/disabling
- [x] Form has correct validation 

NOTE: There is also a fix on this card for auto incrementing visit number. It removes asp-items attribute from an input element. 

fixes #85


src/UDS.Net.Forms/Models/UDS3/D2.cs
Oddvocado, line 1: Good catch on line 88-104!

Oddvocado (approved):
Looks good! The form works as it should.

PR #101, opened 2023-11-08
Add the disable functionality

### LINKED ISSUE #83: C1 Form UI Behavior and Finalization

mlan225:
Resolves: #83 

### List of changes and additions made to the C1 functionality
- [ ] Validation message added to question 1 other field
- [ ] No validation message for 1a1 other field when error, just the red outline
- [ ] 2b other field does not display error message
- [ ] 3a cannot be completed with an input of 88/88/8888 (unknown), but it CAN be partial saved. CAN be accepted on a valid date for completion
- [ ] 3a1 does not follow validation range of the pdf guide (0 - 25) all values are accepted
- [ ] 3b validation is checking the wrong range, won't accept 10 or 20
- [ ] 4a not accepting 10 on its 0-17 range (it accepts 1)
- [ ] 12a not accepting input of 40
- [ ] Question 8b does not save when submitted
- [ ] Add in required if range for optional inputs

### Reviewing the form
The form is pretty straightforward to follow. Optional inputs are required after being enabled, all parent inputs are required to have values on completion. Range inputs enable child fields within valid range. Following the error messages after trying to complete a blank form should give you a roadmap for checking. 

### Required for testing
You will need to run the web PR alongside this PR here: UK-SBCoA/uniform-data-set-dotnet-api#19

### DIFF
--- a/src/UDS.Net.Forms/Models/UDS3/C1.cs
+++ b/src/UDS.Net.Forms/Models/UDS3/C1.cs
@@ -82,6 +82,7 @@ namespace UDS.Net.Forms.Models.UDS3
         public int? LOGIYR { get; set; }
 
         [Display(Name = "Total score from the previous test administration")]
+        [RegularExpression("^(\\d|1\\d|2[0-5]|88)$", ErrorMessage = "(0-25, 88)")]
         [RequiredOnComplete]
         public int? LOGIPREV { get; set; }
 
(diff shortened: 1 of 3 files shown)


smiththay (changes requested):
@mlan225 everything looks good but I did notice that **LOGIPREV** was missing its allowable codes.
Screenshot 2023-11-14 at 11 58 49 AM
Screenshot 2023-11-14 at 12 07 55 PM

mlan225:
Good catch @smiththay, added the regex validation range to LOGIPREV

smiththay (approved):
@mlan225 Nice, good to merge!

Issue #102, opened 2023-11-17
Milestones


ashleybot:
- [ ] Create set of pages to view milestones, add, and edit milestones.

/Pages/Milestones

- [ ] Update participation details page with a html table of milestones (underneath visits).

mlan225:
Hey @ashleybot in the meantime while I build out the views, I wanted to ask about the milestone data that we want to show up on the participations view. 

On the UDS application we have it set up like this:
![Image](https://github.com/UK-SBCoA/uniform-data-set-dotnet-web/assets/17088502/72c2abf4-98a1-4866-88d3-5eb2775ba5f9)

In the milestone model in the api, I don't believe I see properties for the Type (or "kind") and Details. Right now im displaying the minimum available data on the participations view. The Kind property seems to be something that is inherited from the formDto and not the BaseDto that the M1 is inheriting from.

I'll have a better POC of what im currently doing with it after I create the additional views, but I'm keeping in mind that we may want to apply the additional functionality for "miestone type" and "details" as I continue to build things out.

ashleybot:
The **Milestone type** in the screenshot can be determined by which questions the user completes. To summarize, if `DECEASED == 1 || DISCONT == 1` then the "type" is _Change followed by no further contact_, otherwise it is _Change followed by continued contact._

It might help to take a look at the paper form and DED documentation https://naccdata.org/data-collection/forms-documentation/standalone

mlan225:
Ran into and worked through some interesting things last week. Kind of a lot of text here so we can walk through it next time we are able to discuss things: 

- Custom attributes are checking for a type of FormModel, and since we talked about a Milestone not being a "form" and not inheriting from formModel, the custom attributes such as "required if" will need to have an additional file added that works with non formModel types.


![Image](https://github.com/UK-SBCoA/uniform-data-set-dotnet-web/assets/17088502/4e3969da-5e68-4fab-b360-cd45f04c8c6f)


- I had to do a little bit of a workout around when getting a single milestone. As of now there is only a route for getting all milestones within a participation. On the web project, I just had to get the collection of milestones and then further filter in order to have a single milestone for the edit functionality
- Modified the API, but in a way that does not adjust any existing contracts. There was a comment regarding this that you had some previous concerns about and I found to be true, so I adjusted the query/variable pair. The argument for `putMilestone` was initially Id and `milestoneId` instead of `formId` and it was indeed causing the issue stated above in the comment


[...]

PR #103, opened 2023-11-17
Add blank table to participation page

mlan225:
Resolves: #102 

Requires this PR of the uniform data set api: https://github.com/UK-SBCoA/uniform-data-set-dotnet-api/pull/20

### Participation details 
(https://localhost:4811/Participations/Details/id)
- [x] Milestone added to the participation view
- [x] Details for milestones in the participation details table display when 2a is given a value and 2b has at least one checkbox marked
- [x] Can "create milestone" with the create new button
- [x] Can edit milestone with the "edit" link
- [x] Can view milestone details with the "view" link

### Milestone Edit & New views 
(edit link or 'create milestone' button on participation details view) 
- [x] Milestone type of 1  will disable Box B and MilestoneType of 2 will disable Box A
- [x] Validation messages display upon submitting null data when the field is required (submitting a new empty form will give a good idea of the required form fields) 
- [x] Can save Milestone
- [x] Editing milestone loads the correct values


src/UDS.Net.Forms/Pages/Milestones/Create.cshtml.cs
```diff
@@ -0,0 +1,58 @@
...
+
+            Isvalid(Milestone);
+
+            if (ModelState.IsValid)
+            {
+                //TODO sending participationId from page model for now, I think we want to send the Id from visit/participation data?
+                //the api checks to make sure participationId and milestone.ToEntity() participationId are the same
+                await _participationService.AddMilestone(User.Identity?.Name, Milestone.ParticipationId, Milestone.ToEntity());
+
+                return RedirectToPage("/Participations/Details", new { Id = Milestone.ParticipationId });
```
mlan225, line 52: This applies to both the Edit and Create files

The API uses the participation Id to compare it to the participation Id in the milestone model in order to make sure they line up. Right now, I'm just using the `participationId` from the milestone model instead of an id passed into it from elsewhere, so it will always be the same Id when hitting the API from the web app because they are essentially the same property being compared in the end.

My question is if this is okay? i'm assuming the app should always be sending correct data so that is okay, and the comparison check in the API is more for catching inconsistent participation IDs from something that is NOT the uds web app? It seems like i'm cheating the check you implemented.

src/UDS.Net.Forms/Pages/Milestones/Details.cshtml.cs
```diff
@@ -0,0 +1,18 @@
...
+
+namespace UDS.Net.Forms.Pages.Milestones
+{
+	public class DetailsModel : EditModel
+    {
+        public DetailsModel(IParticipationService participationService) : base(participationService)
+        {
+        }
+
+    }
```
mlan225, line 17: The details page model inherits from the edit model in able to use its `GET` method for displaying values. So there are no additional methods needed here.

src/UDS.Net.Forms/Pages/Milestones/Edit.cshtml.cs
```diff
@@ -0,0 +1,66 @@
...
+            //TODO temporary method for getting the selected milestone for edit. Will need to create "getById" in the API
+            var milestonesByParticipationId = await _participationService.GetMilestonesByParticipationId(participationId);
+
+            if(milestonesByParticipationId.Count() < 1)
+            {
+                return NotFound($"No milestones found within participationId of: {participationId}");
+            }
+
+            //TODO will update this functionality with a singular GET for milestone
+            var milestoneFound = milestonesByParticipationId.Where(m => m.Id == id).FirstOrDefault().ToVM();
```
mlan225, line 31: We already talked about updating the API at some point, I'm leaving the TODO comments in as a reminder

src/UDS.Net.Forms/Models/MilestoneModel.cs
```diff
@@ -0,0 +1,72 @@
...
+        public int? DROPREAS { get; set; }
+        public DateTime CreatedAt { get; set; }
+        public string CreatedBy { get; set; }
+        public string? ModifiedBy { get; set; }
+        public string? DeletedBy { get; set; }
+        public bool IsDeleted { get; set; }
+        [Required]
+        [Display(Name = "Which milesone type are you reporting?")]
+        [Range(0, 1)]
+        public int? MilestoneType { get; set; } = null;
```
mlan225, line 69: No longer setting milestone type in the model. Due to some interesting issues I had using this with the behavior, I decided to set the value of the milestone type along with handle the disables related to milestone type within a javascript behavior file `MilestoneTypeBehavior.js`

ashleybot (changes requested):
@mlan225 There are some UI items that need to be addressed

- [ ] Typos
- [ ] Styling
- [ ] Layout (especially for checkboxes)
- [ ] Instructional copy

Some examples:
Screen Shot 2024-01-17 at 8 43 51 AM

Screen Shot 2024-01-17 at 8 52 23 AM

Screen Shot 2024-01-17 at 9 02 49 AM

Screen Shot 2024-01-17 at 9 03 07 AM

mlan225:
Fixed the linter errors

mlan225:
### Posting screenshots of updates to the form and a screenshot of the full form

Screenshot 2024-01-25 at 2 00 51 PM

Screenshot 2024-01-25 at 2 08 01 PM

Screenshot 2024-01-25 at 2 12 24 PM

screencapture-localhost-4811-Milestones-Edit-2024-01-25-14_16_35

mlan225:
Restyled the milestone form layout:
- [ ] Added more visual differentiation to boxes A and B
- [ ] adjusted the appearence of the 5.Death and 6.Dropped sections

screencapture-localhost-4811-Milestones-Create-2024-01-29-09_37_01

mlan225:
Moving into the review, addressed 1 on 1 comments

screencapture-localhost-4811-Milestones-Edit-2024-01-31-10_48_39

src/UDS.Net.Forms/Models/PageModels/MilestonePageModel.cs
```diff
@@ -0,0 +1,272 @@
...
+                        new UIDisableAttribute("Milestone.AUTOPSY"),
+                        new UIEnableAttribute("Milestone.DISCDAY"),
+                        new UIEnableAttribute("Milestone.DISCYR"),
+                        new UIEnableAttribute("Milestone.DROPREAS")
+                    }
+                }
+            }
+        };
+
+        protected private void IsValid(MilestoneModel milestone)
```
mlan225, line 162: Learned a bit about inheritance here:

> Private members are visible only in derived classes that are nested in their base class. Otherwise, they are not visible in derived classes.

This is probably why I had the method as a public initially, so I could get it to work during development and never came back to find out what was going on. In short, because the Edit and create pageModels are not nested in the MilestonePageModel class and are only separate derived classes, then the private methods will not be inherited. 

If you want to inherit a private method in a derived class that is not nested in the base class, then you need to use `protected` or `protected private`
> A protected member is accessible within its class and by derived class instances.

> The private protected keyword combination is a member access modifier. A private protected member is accessible by types derived from the containing class, but only within its containing assembly. (The private protected access modifier is valid in C# version 7.2 and later.)

I used `protected private` here for the extra layer of protection, even though we don't intend to use this method within another project, it cannot be accessed from an external package that inherits this milestonePageModel class.

mlan225:
Something to point out is that because of the client-side validation for the milestoneType on the view, if a user selects `NO FURTHER CONTACT` for milestone type but then forgets to input an option for `deceased` or `withdrawn`

Screenshot 2024-01-31 at 11 51 54 AM

Then when the page reloads to show the validation errors, the milestone type WILL be changed after identifying that `deceased` or `withdrawn` are not selected, and setting it for `continued contact`. This is because when a form is loaded with either a `deceased` or `withdrawn` value, milestone type will need to be set to identity this as `continued contact` because they are neither deceased or withdrawn.

Screenshot 2024-01-31 at 11 52 01 AM

### Steps to reproduce:
1. create a new milestone form
2. set milestone type as  `0. No further contact`
3. Do not check `deceased` or `withdrawn`
4. Submit form and see that milestone type value has been changed and is now disabling box B

src/UDS.Net.Services/Extensions/DtoToDomainMapper.cs
```diff
@@ -2,6 +2,8 @@
 using System.Collections.Generic;
 using System.ComponentModel;
 using System.Linq;
+using System.Net.NetworkInformation;
```
ashleybot, line 5: @mlan225 Looks like the IDE might have incorrectly added some usings here.

src/UDS.Net.Forms/UDS.Net.Forms.csproj
```diff
@@ -26,4 +26,10 @@
 	<ItemGroup>
 		<PackageReference Include="Microsoft.AspNetCore.Mvc.Razor.RuntimeCompilation" Version="6.0.0" />
 	</ItemGroup>
+	<ItemGroup>
```
ashleybot, line 29: @mlan225 Some cleanup in the csproj

ashleybot (approved):
Ready to merge! @mlan225 use the squash-and-merge feature in the GitHub UI this time.

PR #149, opened 2024-03-20
Add pagination to participations

ashleybot:
- [x] You'll need to add more than 10 participation records to your local dev env in order to test
- [x] Paginates participations table in page sizes of 10
- [x] Clicking "Next" button goes to the next page
- [x] Clicking "Previous" button goes to the previous page
- [x] If there is no previous or next page available then the button is disabled
- [x] Pagination area shows calculation of current record indices for the current page including overall total

Screen Shot 2024-03-20 at 1 21 21 PM


Oddvocado (approved):
Looks good! Everything works as described!

PR #159, opened 2024-04-01
Update B9

mlan225:
Resolves: #142


src/UDS.Net.Services/DomainModels/Forms/B9FormFields.cs
```diff
@@ -112,14 +162,22 @@ public B9FormFields(FormDto dto)
...
+        private static int? ConvertBoolToInt(bool? property)
+        {
+            if (property == true) return 1;
+            if (property == false) return 0;
+            return null;
+        }
+
         public string GetDescription()
         {
             return "Clinician Judgment of Symptoms";
```
ashleybot, line 175: @mlan225 typo here

src/UDS.Net.Services/Extensions/DomainToDtoMapper.cs
```diff
@@ -843,6 +844,14 @@ public static B9Dto ToDto(this B9FormFields fields)
         {
             return new B9Dto
             {
+                DECCOG = fields.DECCOG,
+                DECMOT = fields.DECMOT,
+                PSYCHSYM = fields.PSYCHSYM,
+                DECCOGIN = fields.DECCOGIN,
+                DECMOTIN = fields.DECMOTIN,
+                PSYCHSYMIN = fields.PSYCHSYMIN,
+                DECCLIN = ConvertIntToBool(fields.DECCLIN),
```
mlan225, line 853: DECCLIN, DECCLCOG, and DECCLMOT are using a small function to convert them into bool? because the API has them as int? 

However, a similar question with the property DECCLBE is a bool? type in the API

For now I will convert them accordingly to avoid having to update the API. We can change this in the future if we would like, but has no effects on the end result as it is now.

mlan225:
Missing CANNABUSE variable for checkbox 12r1e in the API

image

mlan225:
@ashleybot so the ol' "get to it after I finish up" caught up with me

There is a variable missing from the API that's present in the documentation that I mentioned in a comment. The CANNABUSE

src/UDS.Net.Services/DomainModels/Forms/B9FormFields.cs
mlan225: idk if my eyes just don't work, but I ran the return string through Grammarly like 3 times and double-checked the form name... and somehow I don't see the typo 👀

Oddvocado:
Looks good! Noticed a couple things

- [ ] CANNABUSE is missing UI behaviors 
image

- [ ] If BEAHALL has a value of 1 ("Yes") BEAHSIMP and BEAHCOMP are not required. I didn't see this mentioned as a requirement here --> https://github.com/naccdata/uniform-data-set/blob/main/forms/uds/b9/form_b9_ivp_questions_and_vars.csv but the question does read in a way that would make me think 12i1 and 12i2 would required a response when 12i is marked as "Yes". Either way, thought it was worth mentioning.


image


Other than that everything looks good!

mlan225:
@Oddvocado I think you're totally right with that second one, the sub-questions SHOULD be required. I also added a requirement for the pair above those 2 you pointed out as well.

- [ ] Added CANNABUSE to the behaviors for BESUBAB. CANNABUSE should disable properly now when BESUABAB is 0 or 9
- [ ] BEAHSIMP and BEAHCOMP are required if BEAHALL is 1
- [ ] BEVPATT and BEVWELL are required when BEVHALL is 1

Oddvocado (approved):
Everything looks good!

PR #160, opened 2024-04-10
D1a UI

### LINKED ISSUE #155: Web App: UI for D1a

smiththay:
fixes #155


Oddvocado (changes requested):
Looking good! Ran into a few issues that will need to be addressed. 

- [ ] Question 8 "PREDOMSYN" is missing. I didn't find it in the model nor did I see it in the API. Not sure if it was intentionally omitted but I wanted to point this out.
image
image

- [ ] The subcounters in section 3 are incremented incorrectly
image

- [ ] Section 1 question 2 has formatting issues
image
image

- [ ] Ran into an issue with question 2 "NORMCOG" enabling question 3 "DEMENTED".  When NORMCOG has a value of 0, DEMENTED is not always being enabled. Also had trouble saving the value of DEMENTED when marking the form as complete.
image

smiththay:
@Oddvocado Thanks for those catches, ready for another review

Oddvocado (changes requested):
Looks good! Did notice one issue though. I was able to save the form as complete while selecting more than one diagnosis as primary
tempsnip

smiththay:
@Oddvocado good catch, ready for another review. 

There is an issue that will have a separate card for. Checkboxes that are enabled/disabled by a radio button doesn't save initially if validation on the form kicks on but will save if reentered.

Oddvocado:
@smiththay Looks good! Only thing I noticed was this curly bracket hanging out at the bottom of the form. Other than that looks like it is ready to merge!

image

Issue #162, opened 2024-04-22
Github Super linter update


mlan225:
Update the super linter version and document any errors that result from the update

- In addition to super linter update, upgrade the checkout action as well

mlan225:
## On first check after update to superlinter v6
View full action run here: https://github.com/UK-SBCoA/uniform-data-set-dotnet-web/actions/runs/8786968800/job/24111065743

### CHECKOV healthcheck missing
- Check: CKV_DOCKER_2: "Ensure that HEALTHCHECK instructions have been added to container images"
```
  		1  | #See https://aka.ms/containerfastmode to understand how Visual Studio uses this Dockerfile to build your images for faster debugging.
  		2  | 
  		3  | FROM mcr.microsoft.com/dotnet/aspnet:6.0 AS base
  		4  | WORKDIR /app
  		5  | EXPOSE 82
  		6  | EXPOSE 445
  		7  | 
  		8  | FROM mcr.microsoft.com/dotnet/sdk:6.0 AS build
  		9  | WORKDIR /src
  		10 | COPY ["UDS.Net.Web.MVC/UDS.Net.Web.MVC.csproj", "UDS.Net.Web.MVC/"]
  		11 | COPY ["UDS.Net.Forms/UDS.Net.Forms.csproj", "UDS.Net.Forms/"]
  		12 | COPY ["UDS.Net.Services/UDS.Net.Services.csproj", "UDS.Net.Services/"]
  		13 | COPY ["UDS.Net.Web.MVC.Services/UDS.Net.Web.MVC.Services.csproj", "UDS.Net.Web.MVC.Services/"]
  		14 | RUN dotnet restore "UDS.Net.Web.MVC/UDS.Net.Web.MVC.csproj"
  		15 | COPY . .
  		16 | WORKDIR "/src/UDS.Net.Web.MVC"
  		17 | RUN dotnet build "UDS.Net.Web.MVC.csproj" -c Release -o /app/build
  		18 | 
  		19 | FROM build AS publish
  		20 | RUN dotnet publish "UDS.Net.Web.MVC.csproj" -c Release -o /app/publish /p:UseAppHost=false
  		21 | 
  		22 | FROM base AS final
  		23 | WORKDIR /app
  		24 | COPY --from=publish /app/publish .
  		25 | ENTRYPOINT ["dotnet", "UDS.Net.Web.MVC.dll"]
```

### CHECKOV docker no user created for container
[...]

PR #163, opened 2024-04-22
Update super linter version to 6

mlan225:
Resolves: #162


mlan225:
Checkov, the package that is detecting the errors in the dockerfile and github files can be disabled via a environment variable. This seems like a short term solution to a bigger issue though, so I will attempt to get checkov to skip the github and dockerfile files still. The previous env variables and regex skips did not seems to work so far though.

ashleybot (approved):
Good job @mlan225 !

PR #178, opened 2024-07-10
A5 d2 diabglp1 diabrecact

### LINKED ISSUE #175: A5D2: DIABGLP1, DIABRECACT
Add DIABGLP1, DIABRECACT to UI, model, and mapping.

smiththay:
This will need to be run along slide PR # 67 from the UDS API. This PR adds `diabglp1` and `diabrecact` to the A5D2 UI, model, and mapping.
Fixes #175


mlan225:
@smiththay looks good! Values save in the database and the form completes and partial saves successfully 

I pushed up a commit that only effects the .gitleaksignore file. Since there was code added, it changed the line that the existing ignore was targeting, so checks didn't pass.

PR #184, opened 2024-07-12
A5D2 DEPRTREAT UI Behavior

### LINKED ISSUE #180: A5D2 DEPRTREAT UI Behavior
BUG: DEPRTREAT should be enabled and required only if `MAJORDEP == 1 OR OTHERDEP == 1`

Screenshot 2024-07-11 at 8 28 10 AM

- [x] Fix question numbering (6a1 is repeated twice)
- [x] Only enabled and require DEPRTREAT if MAJORDEP or OTHERDEP = 1

NACC Documentation 

Enable and require if [majordep]='1' or [otherdep]='1'

https://github.com/naccdata/uniform-data-set/blob/main/forms/uds/a5d2/form_a5d2_ivp_questions_and_vars.csv

Oddvocado:
Fixes #180


mlan225:
Talked with @Oddvocado about updating web API package references

mlan225:
Taking a look at the hard-coded question labels

mlan225:
Talked with @Oddvocado about the counter classes, adjusted the hardcoded labels to use this and fixed any following issues in numbering, will pair with Cody to provide some tips on approaching this in the future

mlan225:
@Oddvocado noticed that the validation for DEPRTREAT is halfway implemented. The client side enable/ disable works correctly, but the DEPRTREAT variable is still required when it is disabled

image

mlan225:
Pushed up a commit to help switch the hardcoded labels to the dynamic counter classes, and fixed related issues across the form caused by number adjustments. 

@Oddvocado I remember that you mentioned you'd like me to just push up the solution instead of a full on pairing, so that is viewable in my commit. 

In short though, the form had some issues with the `subsubcounterreset` class was being placed in a div container that was nesting all the numbering within it. I removed the classes in these large parent divs and made sure to only place the `subsubcounterreset` class on the `<tr>` element for the row that needs the subsubcount. Having the large parent divs caused some issues with the resets inside the div.

mlan225:
ALSO! Might need some more info in your PR description 😅

for example, when viewing a PR, I usually try and make sure that the person viewing the PR has all the info they need in order to give a full review of the branch changes, so some things that would be nice to include are: 

- API version the branch uses, IF there was a change to the version (like we had in here)
- A description of some of the changes you've made to the code
- Your own checklist for things that should be checked for a review, things that you want to be confirmed working on your new feature / changes 
- If applicable, some screenshots of working parts of the feature would help illustrate the feature for the reviewer

Oddvocado:
@mlan225 Thanks for catching that! Adjusted the validation so that the form only requires a response when either 6a1 or 6a2 has a value of 1. 

Ready for another pass!

mlan225:
@Oddvocado Noticing that the DEPRTREAT radio value is always saved as 1

### steps to replicate
1. start a new form
2. answer only MAJORDEP and/or OTHERDEP as 1 to enable the DEPRTREAT field
3. partial save the form
4. upon returning, the DEPRTREAT value will be set to 1, the DB value will also be set to 1, even if it is partial saved as 0

image

Oddvocado:
@mlan225 thanks for catching that bug! Turns out there was an issue with DEPRTREAT mapping between the API and the Web project. The API was expecting a Boolean and the web project was expecting an integer. Everything is functioning correctly on my end now. Ready for another review!

src/UDS.Net.Services/DomainModels/Forms/A5D2FormFields.cs
```diff
@@ -350,7 +350,7 @@ public A5D2FormFields(FormDto dto) : this()
                 this.OTHCONDX = a5D2Dto.OTHCONDX;
                 this.MAJORDEP = a5D2Dto.MAJORDEP;
                 this.OTHERDEP = a5D2Dto.OTHERDEP;
-                this.DEPRTREAT = a5D2Dto.DEPRTREAT.HasValue ? 1 : 0;
+                this.DEPRTREAT = a5D2Dto.DEPRTREAT.HasValue ? Convert.ToInt32(a5D2Dto.DEPRTREAT.Value) : (int?)null;
```
ashleybot, line 353: @Oddvocado Probably don't need to add cast to null here `(int?)null`

ashleybot: @Oddvocado Thanks for the learning moment today! You taught me that we do need to have each side of the ternary expression with the same type. .NET Standard 2.1 still uses C# 8.0, so they have evolved this is later versions and that is no longer a requirement, but in our .NET Standard libraries it will be. Thank you @Oddvocado!

Issue #166, opened 2024-05-16
Saving values for checkboxes when enabled/disabled by radio buttons


smiththay:
Checkboxes that are enabled/disabled by a radio button do not save initially if validation on the form kicks on but will save if reentered.

A few forms effected by this are the A5-D2, B9, D1a. 

To replicate:
1.) Go to any form mentioned above or where a radio button is enabled/disabled by a radio button
2.) Must be a new visit and form cannot be in progress or previously saved.
3.) Fill out form and enable checkbox and check it
4.) Any part of the form that is required needs to be left blank so validation will kick on. 
5.) Save/Complete the form
6.) validation will kick on from Q from step 4.
7.) Go to checkbox that was checked, value should be missing.

This issue does not persist if there is no enabling/disabling of the checkbox in the code behind.

mlan225:
### A5D2 checkbox when submitting a completed form
- On the a5d2 that when a form is submitted for the first time and no previous values have been set for a checkbox, the value is removed upon submission and validation identifies it as a missing required value. 

After refilling out the checkbox and submitting the same form, it will be successfully submitted. 

### A5D2 checkbox when submitting a form with an error
 - When submitting a the A5D2 with a validation error and a checkbox value, when the page redirects with the errors included, the checkbox values will be cleared

Looks like the checkbox values are not being sent on the first submission. Looking into a reasoning and resolution

ashleybot:
@mlan225 Noticed that when the parent group changes the enabled/disabled value of the checkboxes. If the parent enabled property changes then the child elements do not correctly get updated. Like the enabled/disabled settings do not propagate through to child elements.

mlan225:
test

mlan225:
Narrowed down the issue even further: 

- When a checkbox can be disabled, if it is changing from a null value (which can happen if it was recently enabled OR a new, unsaved form) the value is not sent along with the post. Something is happening that either does not track that value in the sent request or something is changing it when the form is submitted but *before* the post route is hit 🤔

mlan225:
Investigating the validation.custom javascript

- While I initially felt the issue was somewhere in the behaviors, I checked against removing the the validation javascript files. The only file that was removed and lead to a persisted checkbox "checked" status was the validation.custom. I'm looking around here to see what could be going on in this file

mlan225:
Found the bug in the unobtrusive validation custom file

[...]

PR #186, opened 2024-07-16
Adding some checks to watch the behavior

mlan225:
Resolves: #166 

This branch does not require a specific api version to review as all functionality is tied with web UI interaction, has been updated to run with the main branch of the API at the time of posting

### Adds a check for checkboxes in the unobtrusive.custom.js
- [x] When selecting a parent option that re-enables a checkbox value, the checkbox will now be sent to the controller and saved correctly on the first form post
- [x] Other inputs that can be disabled still function correctly and record values in the db when saving the form: Text, number and radio 
  
### Example with a new A5D2 form
1. checkbox is disabled because the parent value (question 4g) is currently disabling the child checkboxes (4g1)
- image

2. enable checkbox field by selecting option 1 for 4g
- image

3. select a child checkbox value in section 4g1 
- image

4. Submit the form as in progress or complete OR force a validation error in order to revisit the view, the checkbox value should still be selected
 

- [x] You can test this with existing forms as well, as long as the form is loaded with the parent element disabling the checkbox you wish to test for persistence


src/UDS.Net.Forms/wwwroot/js/jquery.unobtrusive.custom_validation.js
```diff
@@ -18,12 +18,13 @@ function setAffect(target, attribute, value) {
...
-        if (element.is(':radio') || element.is(':checked')) {
-          element.removeAttr('checked');
+        // Disable effect by element type
+        if (element.is(':radio') || element.is(':checked') || (element.is(':checkbox'))) {
+            element.removeAttr('checked');
         }
+        //catch all case for all other input types
         else {
-          element.val('');
+            element.val('');
```
mlan225, line 27: If we want further effects with other disable-able inputs in the future, this is where we need to apply them. As written in the original functionality, elements that don't match the condition are simply given a blank value in the else case.

mlan225:
Updated branch with main branch of uniform data set web

smiththay (approved):
@mlan225 selecting a parent option that re-enables a checkbox value saves correctly on the first form post. Tested with in-progress and forced validation. Text, number and radio disable/enable still function correctly and data persist. Tested the function on A5D2, B9 and D1a and all persist correctly. Good fix! Ready to go!

PR #194, opened 2024-07-17
adds UI behavior in code behind and ui attribute

### LINKED ISSUE #128: Milestone type property in VM
Currently the UI does not persist the milestone type on postback so the user would have to reselect type if they do not choose DECEASED or DROPPED on post.

See more here:

> Something to point out is that because of the client-side validation for the milestoneType on the view, if a user selects `NO FURTHER CONTACT` for milestone type but then forgets to input an option for `deceased` or `withdrawn`

Screenshot 2024-01-31 at 11 51 54 AM

> Then when the page reloads to show the validation errors, the milestone type WILL be changed after identifying that `deceased` or `withdrawn` are not selected, and setting it for `continued contact`. This is because when a form is loaded with either a `deceased` or `withdrawn` value, milestone type will need to be set to identity this as `continued contact` because they are neither deceased or withdrawn.

_Originally posted by @mlan225 in https://github.com/UK-SBCoA/uniform-data-set-dotnet-web/issues/103#issuecomment-1919523133_

smiththay:
fixes #128

### DIFF
--- a/src/UDS.Net.Forms/Models/PageModels/MilestonePageModel.cs
+++ b/src/UDS.Net.Forms/Models/PageModels/MilestonePageModel.cs
@@ -159,6 +159,74 @@ namespace UDS.Net.Forms.Models.PageModels
             }
         };
 
+        public Dictionary<string, UIBehavior> MilestoneTypeUIBehavior = new Dictionary<string, UIBehavior>
+        {
+            { "1", new UIBehavior {
+                PropertyAttributes = new List<UIPropertyAttributes>
+                {
+                    new UIEnableAttribute("Milestone.CHANGEMO"),
+                    new UIEnableAttribute("Milestone.CHANGEDY"),
+                    new UIEnableAttribute("Milestone.CHANGEYR"),
+                    new UIEnableAttribute("Milestone.PROTOCOL"),
+                    new UIEnableAttribute("Milestone.ACONSENT"),
+                    new UIEnableAttribute("Milestone.RECOGIM"),
+                    new UIEnableAttribute("Milestone.REPHYILL"),
+                    new UIEnableAttribute("Milestone.REREFUSE"),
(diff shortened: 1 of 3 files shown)


mlan225:
Ran into an interesting error message from the GitHub linter about our javascript file form the Prettier package. It doesn't provide further documentation on the formatting errors, but it tells you the command you need to run using an npm package. The package can run from your local machine and not the project, so it can be used without adding node files to the project.

because I have Node on my machine, I ran the prettier command on the file and pushed up the changes for the test to pass. Other devs may have to do this in the future

smiththay:
@ashleybot Thanks for catching that. When I tested the form for when `MILESTONE TYPE = 1` I entered 99 for the day on that date input so that funky validation didn't kick on. When I went back to test I could see that it was wonky. I got that validation resolved for `CHANGEDY`. I noticed issues for  `ACONSENT` as well so I got the validation resolved for that input. I didn't see any code for validation for questions 3 and 3a so left it alone but if there is some validation needed I will update it. 
Screenshot 2024-08-01 at 9 05 40 AM

If not this is ready for another review.

smiththay:
@ashleybot Im getting failure for package release. I updated API to preview 4.0.1-preview.3 and that includes all the changes that Marvin has made to the A1. So when I update my package references in the Web app to 4.0.1-preview.3, I am missing the definition to the properties that Marvin had taken out for the A1. I believe Marvin has to have his PR for the Web project merged into main, and then I need to merge main into my feature branch to get this package error to pass.

ashleybot (approved):
@smiththay Bug fix looks good! I was able to select `0`, see validation working as expected, save, re-edit and all inputs were in the correct states.

You made some overall improvements along the way too - well done!

Good to update dependency versions and versions for this codebase and merge to main

PR #199, opened 2024-07-18
fix heart attack age property swap

mlan225:
fixes: #198 

Fixed the bug with the incorrect variable HTATTAGE in question 2f1 and noticed that it also impacted the question 2a2, as the properties were switched. I adjusted that issue with both questions. 

- [x] change the incorrect property HTATTAGE and change to PACDEFAGE for question 2f1
- [x] change the invalid property PACDEFAGE to HTATTAGE for question 2a2
- [x] question 2f correctly disables question 2f1 on values 0 and 9 
- [x] question 2a correctly disables question 2a2 on values 0 and 9 

### NACC documentation
image

### PR screenshot
image
image


smiththay (approved):
@mlan225 the updates match NACC documentation and have correct enable/disable behavior as well as validation. Good to go!

PR #216, opened 2024-11-14
removes partial closing tags

### LINKED ISSUE #213: Remove partial closing tags
</partial> is being rendered in resulting page. Often the developer tools extension in browsers will ignore this, so you'll need to look at the true page source to see them.

Image

Remove `</partial>` closing tags. Calls to partials should be self-closing, similar to input or img.

Oddvocado:
Fixes #213 

The goal of this PR is to remove any partial closing tags such as  </partial>. In order to test you will need to inspect the page source rather than using developer tools.


ashleybot (approved):
@Oddvocado I looked at page source and did a search within the source code for good measure - looks good! 👍

PR #218, opened 2024-11-20
Additional a1 validation

### LINKED ISSUE #206: Validation for GENDKN, 	GENNOANS, SEXORNDNK, SEXORNNOAN, SERVED
Image

Missing validation for these 5 properties. [Docs here](https://github.com/naccdata/uniform-data-set/blob/main/forms/uds/a1/form_a1_ivp_error_checks_p.csv)

Oddvocado:
Fixes #206 

This PR adds validation to the following properties:

- GENDKN
- GENNOANS
- SEXORNDNK
- SEXORNNOAN
- SERVED

You can find the relevant NACC documentation [here](https://github.com/naccdata/uniform-data-set/blob/main/forms/uds/a1/form_a1_ivp_error_checks_p.csv)

During testing, I identified a bug that prevented the form from being submitted. The issue arose because Turbo expected a page redirect upon form submission. Although this form does not utilize Turbo Frames, the _LayoutForm partial, shared across all forms, includes the Turbo CDN.

To resolve this, I wrapped the form in an if statement and added the data-turbo="false" attribute to disable Turbo for forms other than C2 and C2T. You can review the commit addressing this fix [here](https://github.com/UK-SBCoA/uniform-data-set-dotnet-web/pull/218/commits/5916d09dbbe05ac0403a4c0f7d6393073a82d1bd)

To enable or disable properties dynamically, I used the [checkboxDisable_controller](https://github.com/UK-SBCoA/uniform-data-set-dotnet-web/blob/main/src/UDS.Net.Forms/wwwroot/js/js_controllers/checkboxDisable_controller.js). However, I had to modify its functionality.

When testing, it would be best to verify that the controller's functionality remains intact across other forms where it is implemented. It appears the A1a form is the only other form utilizing this controller.

### DIFF
--- a/src/UDS.Net.Forms/Models/UDS4/A1.cs
+++ b/src/UDS.Net.Forms/Models/UDS4/A1.cs
@@ -379,7 +379,32 @@ namespace UDS.Net.Forms.Models.UDS4
                 else return null;
             }
         }
-
+        [RequiredOnFinalized(ErrorMessage = "If response 4f is checked then responses 4h and 4i must be blank")]
+        [NotMapped]
+        public bool? TwoSpiritGender
+        {
+            get
+            {
+                if (GENTWOSPIR == true && (GENDKN == true || GENNOANS == true))
+                {
+                    return null;
+                }
+                else return true;
+            }
+        }
+        [RequiredOnFinalized(ErrorMessage = "If response 7a is checked then responses 7f and 7g must be blank")]
+        [NotMapped]
+        public bool? TwoSpiritSexOrientation
+        {
+            get
+            {
+                if (SEXORNTWOS == true && (SEXORNDNK == true || GENNOANS == SEXORNNOAN))
+                {
+                    return null;
+                }
+                else return true;
+            }
+        }
         [Display(Name = "What is your primary language?")]
         [RequiredOnFinalized]
         public int? PREDOMLAN { get; set; }
--- a/src/UDS.Net.Forms/Models/UDS4/A1.cs
+++ b/src/UDS.Net.Forms/Models/UDS4/A1.cs
@@ -431,6 +431,20 @@ namespace UDS.Net.Forms.Models.UDS4
                 else return true;
             }
         }
+
+        [RequiredOnFinalized(ErrorMessage = "If question 16 ('MEDVA') is marked as 'Yes' then question 15 ('SERVED') cannot equal 'Don't Know'")]
+        [NotMapped]
+        public bool? VAMedCare
+        {
+            get
+            {
+                if (MEDVA == 1 && SERVED == 9)
+                {
+                    return null;
+                }
+                else return true;
+            }
+        }
         [Display(Name = "What is your primary language?")]
         [RequiredOnFinalized]
         public int? PREDOMLAN { get; set; }
--- a/src/UDS.Net.Forms/Models/UDS4/A1.cs
+++ b/src/UDS.Net.Forms/Models/UDS4/A1.cs
@@ -364,6 +364,22 @@ namespace UDS.Net.Forms.Models.UDS4
             }
         }
 
+        [RequiredOnFinalized(ErrorMessage = "Both question 15(A1.SERVED) and 16(A1.MEDVA) cannot have a value of \"Don't Know\"")]
+        [NotMapped]
+        public bool? ServedAndVaCare
+        {
+            get
+            {
+                if (SERVED == 9 && MEDVA != 9 ||
+                    SERVED == 1 && (MEDVA == 9 || MEDVA == 1 || MEDVA == 0) ||
+                    SERVED == 0 && MEDVA == null)
+                {
+                    return true;
+                }
+                else return null;
+            }
+        }
+
         [Display(Name = "What is your primary language?")]
         [RequiredOnFinalized]
         public int? PREDOMLAN { get; set; }
--- a/src/UDS.Net.Forms/Models/UDS4/A1.cs
+++ b/src/UDS.Net.Forms/Models/UDS4/A1.cs
@@ -405,32 +405,6 @@ namespace UDS.Net.Forms.Models.UDS4
                 else return true;
             }
         }
-        [RequiredOnFinalized(ErrorMessage = "GENOTHX must be blank when response 4g. is unchecked")]
-        [NotMapped]
-        public bool? GenderSpecify
-        {
-            get
-            {
-                if (!string.IsNullOrWhiteSpace(GENOTHX) && GENOTH == false)
-                {
-                    return null;
-                }
-                else return true;
-            }
-        }
-        [RequiredOnFinalized(ErrorMessage = "SEXORNOTHX must be blank when response 7e. is unchecked")]
-        [NotMapped]
-        public bool? SexOrientationSpecify
-        {
-            get
-            {
-                if (!string.IsNullOrWhiteSpace(SEXORNOTHX) && SEXORNOTH == false)
-                {
-                    return null;
-                }
-                else return true;
-            }
-        }
 
         [RequiredOnFinalized(ErrorMessage = "If question 16 ('MEDVA') is marked as 'Yes' then question 15 ('SERVED') cannot equal 'Don't Know'")]
         [NotMapped]
--- a/src/UDS.Net.Forms/Models/UDS4/A1.cs
+++ b/src/UDS.Net.Forms/Models/UDS4/A1.cs
@@ -364,21 +364,23 @@ namespace UDS.Net.Forms.Models.UDS4
             }
         }
 
-        [RequiredOnFinalized(ErrorMessage = "Both question 15(A1.SERVED) and 16(A1.MEDVA) cannot have a value of \"Don't Know\"")]
+        [RequiredOnFinalized(ErrorMessage = "If Q16. MEDVA (VA medical care) = 1 (Yes) then Q15. SERVED (served in armed forces) should not equal 9 (Don't know),it should be equal to 1 (Yes)")]
         [NotMapped]
         public bool? ServedAndVaCare
         {
             get
             {
-                if (SERVED == 9 && MEDVA != 9 ||
-                    SERVED == 1 && (MEDVA == 9 || MEDVA == 1 || MEDVA == 0) ||
-                    SERVED == 0 && MEDVA == null)
+                if (SERVED == 9)
                 {
-                    return true;
-                }
-                else return null;
+                    if (MEDVA == 1)
+                    {
+                        return null;
+                    }
+                }    
+                return true;
             }
         }
+
         [RequiredOnFinalized(ErrorMessage = "If response 4f is checked then responses 4h and 4i must be blank")]
         [NotMapped]
         public bool? TwoSpiritGender
@@ -406,19 +408,6 @@ namespace UDS.Net.Forms.Models.UDS4
             }
         }
 
-        [RequiredOnFinalized(ErrorMessage = "If question 16 ('MEDVA') is marked as 'Yes' then question 15 ('SERVED') cannot equal 'Don't Know'")]
-        [NotMapped]
-        public bool? VAMedCare
-        {
-            get
-            {
-                if (MEDVA == 1 && SERVED == 9)
-                {
-                    return null;
-                }
-                else return true;
-            }
-        }
         [Display(Name = "What is your primary language?")]
         [RequiredOnFinalized]
         public int? PREDOMLAN { get; set; }
--- a/src/UDS.Net.Forms/Models/UDS4/A1.cs
+++ b/src/UDS.Net.Forms/Models/UDS4/A1.cs
@@ -394,7 +394,7 @@ namespace UDS.Net.Forms.Models.UDS4
                 else return true;
             }
         }
-        [RequiredOnFinalized(ErrorMessage = "If response 7a is checked then responses 7f and 7g must be blank")]
+        [RequiredOnFinalized(ErrorMessage = "If response 7d is checked then responses 7f and 7g must be blank")]
         [NotMapped]
         public bool? TwoSpiritSexOrientation
         {
--- a/src/UDS.Net.Forms/Models/UDS4/A1.cs
+++ b/src/UDS.Net.Forms/Models/UDS4/A1.cs
@@ -405,6 +405,32 @@ namespace UDS.Net.Forms.Models.UDS4
                 else return true;
             }
         }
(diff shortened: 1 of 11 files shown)


mlan225:
- [ ] GENDKN (Don't know) checkbox disables all gender options
- [ ] GENNOANS (prefer not to answer) checkbox disables all gender options
- [ ] SEXORNDNK (Don't know) checkbox disables the other sexual orientation options
- [ ] SEXORNNOAN (prefer not to answer) checkbox disables the other orientation options
- [ ] SERVED radio group cannot be unknown and MEDVA be checked as Yes

It seems like I am able to submit SERVED as 9 (don't know) and MEDVA as 1 (Yes), is this correct? Here are the rules in the docs `If MEDVA=1 then SERVED should not equal 9`

Oddvocado:
@mlan225 thanks for catching that! Added additional validation to the model to prevent form submission when SERVED has a value of 9 and MEDVA has a value of 1

mlan225:
Looks like the validation is catching for the MEDVA and SERVED variables in regard to the listed instructions on the parent issue card 👍 

- [x]  SERVED radio group cannot be unknown and MEDVA be checked as Yes

Looks good to me! Checked to make sure that the A1a that also uses the checkbox disable still functions properly and seems it still works fine. Cody explained the checkbox javascript enhances he made and it all makes sense to me. 

@ashleybot I'll pass the second check on to you

src/UDS.Net.Forms/Models/UDS4/A1.cs
```diff
@@ -364,6 +364,87 @@ public bool? SexualOrientstionIndicated
...
+                if (!string.IsNullOrWhiteSpace(GENOTHX) && GENOTH == false)
+                {
+                    return null;
+                }
+                else return true;
+            }
+        }
+        [RequiredOnFinalized(ErrorMessage = "SEXORNOTHX must be blank when response 7e. is unchecked")]
+        [NotMapped]
+        public bool? SexOrientationSpecify
```
ashleybot, line 423: @Oddvocado Same question for `SEXORNOTH` and `SEXORNOTHX`. If UI behaviors are working correctly, the user should not be able to perform this invalid action, right?

src/UDS.Net.Forms/Models/UDS4/A1.cs
```diff
@@ -364,6 +364,87 @@ public bool? SexualOrientstionIndicated
             }
         }
 
+        [RequiredOnFinalized(ErrorMessage = "Both question 15(A1.SERVED) and 16(A1.MEDVA) cannot have a value of \"Don't Know\"")]
+        [NotMapped]
+        public bool? ServedAndVaCare
+        {
+            get
+            {
+                if (SERVED == 9 && MEDVA != 9 ||
```
ashleybot, line 373: @Oddvocado This conditional statement is complex. Can you make it more readable, maybe by adding more parentheses or nesting the statements?

src/UDS.Net.Forms/Models/UDS4/A1.cs
```diff
@@ -364,6 +364,87 @@ public bool? SexualOrientstionIndicated
...
+                {
+                    return null;
+                }
+                else return true;
+            }
+        }
+
+        [RequiredOnFinalized(ErrorMessage = "If question 16 ('MEDVA') is marked as 'Yes' then question 15 ('SERVED') cannot equal 'Don't Know'")]
+        [NotMapped]
+        public bool? VAMedCare
```
ashleybot, line 437: @Oddvocado Is `VAMedCare` already handled by `ServedAndVaCare`?

ashleybot (changes requested):
@Oddvocado I added a commit to add `asp-validator-for` spans close to the targeted inputs. 

And I asked some questions on different lines. Some states I'm not sure can be done unless UI Behaviors are incorrect, and some things might just be me incorrectly interpreting the spec.

src/UDS.Net.Forms/Models/UDS4/A1.cs
ashleybot: Or could the conditional be added there since that property is already comparing the two?

src/UDS.Net.Forms/Models/UDS4/A1.cs
```diff
@@ -364,6 +364,87 @@ public bool? SexualOrientstionIndicated
...
+                if (SEXORNTWOS == true && (SEXORNDNK == true || SEXORNNOAN == true))
+                {
+                    return null;
+                }
+                else return true;
+            }
+        }
+        [RequiredOnFinalized(ErrorMessage = "GENOTHX must be blank when response 4g. is unchecked")]
+        [NotMapped]
+        public bool? GenderSpecify
```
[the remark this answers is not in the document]
Oddvocado: Went ahead an removed these! I think I added them at some point due to some issues I had with javascript? Can't quite remember anymore. You're correct the this validation is handled by the UIBehaviors

src/UDS.Net.Forms/Models/UDS4/A1.cs
Oddvocado: I don't know what I was thinking with this one haha! I have gone ahead and made it much much simpler.

src/UDS.Net.Forms/Models/UDS4/A1.cs
Oddvocado: Went ahead and removed this! The validation scenario is handled by my most recent tag helper commit

src/UDS.Net.Forms/Models/UDS4/A1.cs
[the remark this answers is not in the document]
Oddvocado: I take that back! I figured out why I did this. If you check and then uncheck "dont know" or "prefer not to answer" it will enable GENOTHX. The user could then potentially select another checkbox while also adding a value to GENOTHX. Checked the database and it does incorrectly save a value for GENOTHX in this sitation.

image

src/UDS.Net.Forms/Models/UDS4/A1.cs
Oddvocado: Same situation here

image

src/UDS.Net.Forms/Models/UDS4/A1.cs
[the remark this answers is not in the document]
ashleybot: @Oddvocado This is sounding like a bug on the other checkboxes' UI behaviors. It looks like there is a data_checkBoxDisable stimulus controller handling that.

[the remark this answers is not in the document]
ashleybot: @Oddvocado Yep, it is a bug with UI behaviors on `GENDKN` and `GENNOANS`. I'll commit the fix for this grouping. Can you make the fix for the other grouping?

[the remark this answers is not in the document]
Oddvocado: Yep!

ashleybot (changes requested):
@Oddvocado Still some more to implement to follow the specifications outlined in the card.

### If one gender is known, cannot check Don't know or No answer

So, UI Behaviors needs to be added to all these checkboxes to disable GENDKN and GENNOANS  as well. Either `data-checkBoxDisable` or `data-affects` could be used for this scenario:

- [ ] GENMAN
- [ ] GENWOMAN
- [ ] GENTRMAN
- [ ] GENTRWOMAN
- [ ] GENNONBI
- [ ] GENTWOSPIR

### If one sexual orientation is known, cannot check Don't know or No answer

Similar situation here. These all need to be updated to disable SEXORNDNK and SEXORNNOAN

- [ ] SEXORNGAY
- [ ] SEXORNHET
- [ ] SEXORNBI
- [ ] SEXORNTWOS

src/UDS.Net.Forms/Pages/UDS4/A1.cshtml
```diff
@@ -718,7 +721,17 @@
                             <div class="relative flex items-start">
                                 <span class="subcounter mr-3 block text-sm font-medium leading-6 text-gray-900"></span>
                                 <div class="flex h-6 items-center">
-                                    @Html.CheckBox("A1.GENOTH", Model.A1.GENOTH, new { @class = "h-4 w-4 rounded border-gray-300 text-indigo-600 focus:ring-indigo-600", data_affects = "true", data_affects_toggle_targets = "[ \"A1.GENOTHX\" ]" })
+                                    <!--// GENOTH is pretty complex because it needs to enable and also disable //-->
+                                    @Html.CheckBox("A1.GENOTH", Model.A1.GENOTH, new
+                                    {
+                                        @class = "h-4 w-4 rounded border-gray-300 text-indigo-600 focus:ring-indigo-600 disabled:bg-slate-50 disabled:text-slate-500 disabled:border-slate-200 disabled:shadow-none", 
+                                        data_affects = "true", data_affects_toggle_targets = "[ \"A1.GENOTHX\"]",
```
ashleybot, line 728: @Oddvocado These UI behaviors are now completely handled by `data-checkBoxDisable` so `data-affects` should be redundant here, right?

src/UDS.Net.Forms/Pages/UDS4/A1.cshtml
```diff
@@ -822,7 +852,17 @@
                             <div class="relative flex items-start">
                                 <span class="subcounter mr-3 block text-sm font-medium leading-6 text-gray-900"></span>
                                 <div class="flex h-6 items-center">
-                                    @Html.CheckBox("A1.SEXORNOTH", Model.A1.SEXORNOTH, new { @class = "h-4 w-4 rounded border-gray-300 text-indigo-600 focus:ring-indigo-600", data_affects = "true", data_affects_toggle_targets = "[ \"A1.SEXORNOTHX\" ]" })
+                                    @Html.CheckBox("A1.SEXORNOTH", Model.A1.SEXORNOTH, new 
+                                    { 
+                                        @class = "h-4 w-4 rounded border-gray-300 text-indigo-600 focus:ring-indigo-600 disabled:bg-slate-50 disabled:text-slate-500 disabled:border-slate-200 disabled:shadow-none", 
+                                        data_affects = "true", data_affects_toggle_targets = "[ \"A1.SEXORNOTHX\" ]",
```
ashleybot, line 858: @Oddvocado Same thing here. There shouldn't be a scenario where we would use both of these `data-affects` and `data-checkBoxDisable`.

src/UDS.Net.Forms/Pages/UDS4/A1.cshtml
Oddvocado: When I removed the data-affects some inputs were being incorrectly enabled when initially loading the form

Oddvocado:
> @Oddvocado Still some more to implement to follow the specifications outlined in the card.
> 
> ### If one gender is known, cannot check Don't know or No answer
> So, UI Behaviors needs to be added to all these checkboxes to disable GENDKN and GENNOANS as well. Either `data-checkBoxDisable` or `data-affects` could be used for this scenario:
> 
> * [ ]  GENMAN
> * [ ]  GENWOMAN
> * [ ]  GENTRMAN
> * [ ]  GENTRWOMAN
> * [ ]  GENNONBI
> * [ ]  GENTWOSPIR
> 
> ### If one sexual orientation is known, cannot check Don't know or No answer
> Similar situation here. These all need to be updated to disable SEXORNDNK and SEXORNNOAN
> 
> * [ ]  SEXORNGAY
> * [ ]  SEXORNHET
> * [ ]  SEXORNBI
> * [ ]  SEXORNTWOS



We discussed this issue over voice chat but wanted to document it for future reference. Currently, the checkboxDisable functionality correctly disables the appropriate group targets. However, a bug arises when multiple checkboxes target the same checkbox group. This results in unexpected behavior where a checkbox may be incorrectly enabled.

For example, in the case of checkboxes 4a and 4b:

When checked, both 4a and 4b correctly disable checkboxes 4h and 4i.

image


The issue occurs when either 4a or 4b is unchecked:

If one of these checkboxes is unchecked while the other remains checked, both 4h and 4i are incorrectly enabled.

image

src/UDS.Net.Forms/Pages/UDS4/A1.cshtml
ashleybot: @Oddvocado If it isn't loading properly then this means that data-checkBoxDisabled is missing some page onLoad functionality. So, that should be fixed here with this PR.

src/UDS.Net.Forms/wwwroot/js/js_controllers/checkboxDisable_controller.js
```diff
@@ -25,44 +26,82 @@ export default class extends Controller {
...
 
-  OnLoad() {
-    this.checkboxTriggerTargets.forEach(() => {
 
-      let disableString = this.checkboxTriggerTarget.dataset.checkboxdisableDisableParam
-      let group = this.checkboxTriggerTarget.dataset.checkboxdisableGroupParam
 
+  }
+
+  OnLoad() {
```
ashleybot, line 82: @Oddvocado Yep, there must be a bug with the on load functionality in checkboxDisable. A1a throws an error when `EXPNOTAPP` is true.

Screenshot 2025-01-08 at 8 32 16 AM

Oddvocado:
Thanks for catching that bug! Some variables in the OnLoad() method had inconsistent formatting, but those issues should now be resolved.

However, there was still unexpected behavior on the initial page load, particularly when the data-affects tag was removed. For example, if GENOTH was unchecked and the data-affects tag was removed, GENOTHX would incorrectly be enabled on page load. This happened because the enabling and disabling logic on page load only checked for targets that were already checked, from this line of code

```   
if (checkbox.checked) {
     this.ToggleGroup({ params: { enablegroup, disablegroup, togglestate }, currentTarget: checkbox })
   } 
 ```
   
Removing this line of code introduced other issues. For instance, if GENDKN and GENNOANS both disable questions 4a–4g when checked, questions 4a–4g would be incorrectly enabled if GENDKN was checked first and GENNOANS was subsequently disabled. This occurred because the controller triggered the 'ToggleGroup' method for both GENDKN and GENNOANS whether they were checked or not.
  
  
image


Long story short, I added an additional parameter and method to handle the logic. These changes allowed me to remove the data-affects tag, and everything now appears to be functioning as expected!

ashleybot (approved):
@Oddvocado and @ashleybot paired and realized that the most tricky part of these checkbox behaviors is that there are two "parent" checkboxes controlling the same group of checkboxes. So, its a race condition situation. We handled the scenario by checking if the other checkbox is disabled then it doesn't control anything.

PR #225, opened 2024-12-12
 Maintain disabled state in RMREAS, RMMODE, NOT on postback

### LINKED ISSUE #220: Maintain disabled state in RMREAS, RMMODE, NOT on postback
Some forms do not allow a `MODE` == remote or not completed. In these scenarios, subsequent select inputs are correctly rendered as disabled, and with appropriate select items.

For example, the **B1** does not allow remote administration. So, the `RMREAS` and `RMMODE` select inputs are disabled and only have one item with a null value. In this way the form is posted with the correct values every time.

Image

Another example, the **B4** is required and **must** be completed. So, the `NOT` select input is disabled and only has one item with a null value.

Image

When a form is submitted and errors are present, the model is returned with validation errors. Subsequent form submissions aren't broken, so this isn't a bug, but we can improve the feel of the UX by maintaining the disabled state of the select inputs.

Here is the B1 after postback with validation errors. `RMREAS` and `RMMODE` are _incorrectly_ enabled.

Image

Here is the B4 after postback with validation errors and `NOT` is _incorrectly_ enabled.
 Image

This change has a wide breadth of effect (all forms use the partial) so extensive testing will need to be conducted before PR review.

Oddvocado:
Fixes #220 

This update addresses an issue where the RMREAS, RMMODE, and NOT properties were incorrectly enabled when a form was submitted as "Finalized." To resolve this, a CSS class named never-enabled was added. This class overrides the valid class applied by jQuery validation during an HTTP POST request.

To test this fix:
- [x] Submit a form containing validation errors.
- [x] When the page reloads with errors displayed, ensure that the RMREAS, RMMODE, and NOT properties remain in their disabled state.

image

All forms use this partial, so multiple forms will need to be tested during review to ensure consistent behavior.

### DIFF
--- a/src/UDS.Net.Forms.Tests/UDS.Net.Forms.Tests.csproj
+++ b/src/UDS.Net.Forms.Tests/UDS.Net.Forms.Tests.csproj
@@ -5,7 +5,7 @@
 		<ImplicitUsings>enable</ImplicitUsings>
 		<Nullable>enable</Nullable>
 		<IsPackable>false</IsPackable>
-		<ReleaseVersion>4.6.3</ReleaseVersion>
+		<ReleaseVersion>4.6.4</ReleaseVersion>
 	</PropertyGroup>
 	<ItemGroup>
 		<PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.11.1" />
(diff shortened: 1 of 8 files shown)


ashleybot (approved):
@Oddvocado Tested both forms where the dropdown are enabled (remote is allowed) and disabled (remote is not allowed)

Also tested when form is required and not completed reason is disabled.

All scenarios follow the spec outlined in the card!

- [x] A1 (remote allowed, optional)
- [ ] B1 (remote not allowed, optional)

Issue #221, opened 2024-12-02
C2 MOCA Delayed Recall Enhanced Validation


ashleybot:
Enhance C2 with validation for all permutations (shown in table reference below).

Known rules:

- If MOCARECN != 88 && MOCARECC != 88 && MOCARECR != 88 then MOCARECN + MOCARECC + MOCARECR <= 5
- If MOCARECN == 5 then MOCARECC should equal 88 and MOCARECR should equal 88
- If MOCARECN + MOCARECC == 5 then MOCARECR should equal 88
- Study the table reference, there may be more rules

**Table reference**

![c2stuff](https://github.com/user-attachments/assets/6de50f62-6a24-4213-b08e-81817bcebb7e)

mlan225:
Looks like the value MOCARECN (question 1t.) only allows for values  0 - 5 and 95 - 98, however in the first instruction on this card it mentions if this value is not equal to 88. Has there been any change in valid values for this question or was this part of your inference from the table above?

![Image](https://github.com/user-attachments/assets/16c62605-eeed-4357-9b75-6b91df20e038)

mlan225:
i'll make the adjustment to that ruling in the comment above to account for the 95 - 98 values

ashleybot:
Official docs =>

https://github.com/naccdata/uniform-data-set/blob/main/forms/uds/c2/form_c2_ivp_questions_and_vars.csv

-- | -- | -- | -- | -- | -- | -- | -- | --
c2 | I | 1t. Delayed recall -- No cue | MOCARECN | Conditional (e.g. skip pattern) | Integers 0-5, 95-98 | 0, 0\|1, 1\|2, 2\|3, 3\|4, 4\|5, 5\|95, Physical problem\|96, Cognitive/behavior problem\|97, Other problem\|98, Verbal refusal | Integer | [mocacomp]='1'
-- | -- | -- | -- | -- | -- | -- | -- | --
30 | c2 | I | 1u. Delayed recall -- Category cue | MOCARECC | Conditional (e.g. skip pattern) | Integers 0-5, 88 | 0, 0\|1, 1\|2, 2\|3, 3\|4, 4\|5, 5\|88, Not applicable | Integer | [mocacomp]='1'
[...]

PR #231, opened 2024-12-18
add rules and adjust existing rule

mlan225:
FIxes: #221 

These questions are the focus of the PR. The numbering can change depending on the form types, but the labels and validation will remain the same.

image

example of error text display

image

## Adjust the validation for the C2 in all versions of the form (InPerson and remote): 
- [ ] For Delayed recall questions: 'No cue', 'Category cue', and 'Recognition' the sum of questions with a value of 0 - 5 should be less than or equal to 5
- [ ] If Delayed recall - No cue is equal to 5, then Delayed recall - Category cue and Delayed recall - Recognition should both be 88
- [ ] If the sum of Delayed recall - No cue and Delayed recall - Category are equal to 5, then Delayed recall - Recognition should be 88

### DIFF
--- a/src/UDS.Net.Forms/Models/UDS4/C2.cs
+++ b/src/UDS.Net.Forms/Models/UDS4/C2.cs
@@ -671,22 +671,24 @@ namespace UDS.Net.Forms.Models.UDS4
         }
 
         //non mapped property MOCARECN =! 88 && MOCARECC =! 88 && MOCARECR != 88 then all 3 properties should be <= 5
-        [RequiredOnFinalized(ErrorMessage = "If questions 1t, 1u, and 1v are NOT 88, then the sum of these 3 should be less than or equal to 5")]
+        [RequiredOnFinalized(ErrorMessage = "For Delayed recall questions: 'No cue', 'Category cue', and 'Recognition' the sum of questions with a value of 0 - 5 should be less than or equal to 5")]
         [NotMapped]
         public bool? DelayedRecallValidSum
         {
             get
             {
-                if (MOCARECN != 88 && MOCARECC != 88 && MOCARECR != 88)
-                {
-                    if (MOCARECN + MOCARECC + MOCARECR > 5) return null;
-                }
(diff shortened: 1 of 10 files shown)


mlan225:
Josh and I discovered that there is an extra case that delayed recall questions can be: 
1. value
2. 88
3. value

With this in mind, we still needed to check if all of the values that were NOT 88 or 93 - 95 did not have a sum of > 5

src/UDS.Net.Forms/Models/UDS4/C2.cs
```diff
@@ -674,6 +670,68 @@ public bool? RESPVALReasonIndicated
...
+                }
+
+                if (MOCARECR != null)
+                {
+                    mocarecrValue = MOCARECR == 88 ? 0 : MOCARECR;
+                }
+
+                if (mocarecnValue + mocareccValue + mocarecrValue > 5) return null;
+
+                return true;
```
mlan225, line 701: Made this code null safe, even though it *should* not be possible to reach a finalized state AND have null values for MOCARECN, MOCARECC, and MOCARECR.

ashleybot (approved):
@mlan225 Looks good! 👍  I added a small commit so the display of the error messages follows the existing behavior in the form (displaying in the 3rd column).

### Existing layout of error messages in the 3rd column:
Screenshot 2024-12-20 at 8 25 40 AM

### Failure states
`MOCARECN + MOCARECC + MOCARECR > 5`
Screenshot 2024-12-20 at 8 26 19 AM

`If MOCARECN == 5 then ( MOCARECC != 88 && MOCARECR != 88)`
Screenshot 2024-12-20 at 8 27 50 AM

### Success states
`If MOCARECN == 5 then ( MOCARECC == 88 && MOCARECR == 88)`
Screenshot 2024-12-20 at 8 28 18 AM

`MOCARECN + MOCARECC + MOCARECR <= 5`
Screenshot 2024-12-20 at 8 29 15 AM

`if MOCARECN == 95 then MOCARECC + MOCARECR <= 5`
Screenshot 2024-12-20 at 8 28 48 AM

PR #235, opened 2025-01-09
 D1a Normal Cognition Section 3 validation bug

### LINKED ISSUE #234: D1a Normal Cognition Section 3 validation bug
Steps to replicate:

1. Create a new visit (or use existing)
2. Create a new D1a (or use existing)
3. Select `2` for `DXMETHOD`
4. Select `1` for `NORMCOG`
5. Select `0` for `SCD`
6. Change Save status to _Finalized_
7. Validation _incorrectly_ requires Section 3 to have one or more syndromes present and will not allow finalization

Image

Correct behavior:

When NORMCOD = 1 and SCD = 0 the form should be able to be finalized without a syndrome being selected in section 3.

Oddvocado:
Fixes #234 

This bug prevented form submission when either question 2a or 2b was answered in a way that ended the form. Validation was incorrectly requiring section 3 to be completed, even though it was irrelevant in this case.

image

The issue stemmed from an unmapped Boolean property in section 3, which triggered a model error if no responses were recorded. This property did not account for cases where the form should end early based on responses to 2a or 2b.

To fix this, I updated the model class by adding conditions to handle these scenarios properly.

### DIFF
--- a/src/UDS.Net.Forms.Tests/UDS.Net.Forms.Tests.csproj
+++ b/src/UDS.Net.Forms.Tests/UDS.Net.Forms.Tests.csproj
@@ -5,7 +5,7 @@
 		<ImplicitUsings>enable</ImplicitUsings>
 		<Nullable>enable</Nullable>
 		<IsPackable>false</IsPackable>
-		<ReleaseVersion>4.6.5</ReleaseVersion>
+		<ReleaseVersion>4.6.6</ReleaseVersion>
 	</PropertyGroup>
 	<ItemGroup>
 		<PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.11.1" />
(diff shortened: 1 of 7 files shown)


ashleybot (approved):
@Oddvocado Nice work!

- [x] Able to successfully finalize if NORMCOG = 1
- [x] Validation correctly prevents finalization if NORMCOG = 0 and nothing from section 3 is checked
- [x] Validation correctly prevents finalization if NORMCOG = 0 and if something is checked, but diagnosis is not selected (primary, contributing, non-contributing)
- [ ] Validation does correctly prevents finalization if NORMCOG = 0 and more than one condition is indicated as a primary diagnosis 

Good to bump to version 4.6.6. For this one, also update `UDS.Net.Forms `and `UDS.Net.Services` dependencies in COA.UDSv4 and make a hotfix commit there.

Issue #241, opened 2025-01-13
A5D2 PACDEFAGE requirement bug


mlan225:
## PACDEFAGE bug in A5D2
Form is incorrectly requiring a PACDEFAGE value when the conditions for a null value have been met

![Image](https://github.com/user-attachments/assets/15d7784e-a1f8-4495-9b35-ace9df2bfe0c)

The issue seems to be coming a bug in the required if range attribute incorrectly targeting HRTATTACK

![Image](https://github.com/user-attachments/assets/a2a33f1f-9533-4cf1-9bc4-f18bb050e2cb)

Steps for fix: 
- [ ] Adjust the requirement for the RequiredOnFinalize attribute in the A5D2 for PACDEFAGE. This field should require `CVPACDEF` to be 1 or 2 to expect a value, and 0 or 9 to allow null

PR #242, opened 2025-01-13
Adjust the requiredIfRange value for PACDEFAGE

mlan225:
Fixes: #241 

### Adjust requirement for PACDEFAGE to address validation bug in form A5D2
Adjust the requirement for PACDEFAGE from HRTATTACK to CVPACDEF. Due to the razor class library being used in UDS4, this bug also effects the UDS4 application

- [ ] Form should should now save when question 2f is 1 or 2 and a valid input for 2f1 is present OR 2f is 0 or 9 and question 2f is null
image

### DIFF
--- a/src/UDS.Net.Forms.Tests/UDS.Net.Forms.Tests.csproj
+++ b/src/UDS.Net.Forms.Tests/UDS.Net.Forms.Tests.csproj
@@ -5,7 +5,7 @@
 		<ImplicitUsings>enable</ImplicitUsings>
 		<Nullable>enable</Nullable>
 		<IsPackable>false</IsPackable>
-		<ReleaseVersion>4.6.7</ReleaseVersion>
+		<ReleaseVersion>4.6.8</ReleaseVersion>
 	</PropertyGroup>
 	<ItemGroup>
 		<PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.11.1" />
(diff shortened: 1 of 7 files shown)


ashleybot (changes requested):
@mlan225 

- [x] Bug with `PACDEFAGE` incorrectly required when `HRTATTACK` is filled out has been fixed.

- [ ] Additionally, `PACDEFAGE` now correctly required when `CVPACDEF` within and validation correctly stops form finalization.

Screenshot 2025-01-13 at 4 18 54 PM

## Needs some more work

However, there are other bugs made obvious now with HRTATTACK that need to be fixed along with this.

- [ ] `HRTTAGE` should be required when `HRTATTACK` is 1 or 2

Screenshot 2025-01-13 at 4 16 55 PM

mlan225:
Good catch @ashleybot 👍 , I updated the validation for question 2a2 to correctly rely on 2a being 1 or 2

### Update validation for question 2a2 to rely on 2a = 1 || 2 
- [ ] When question 2a is 1 || 2 , a validation error will trigger if question 2a2 is unanswered
- [ ] When question 2a is 0 || 9, the form can successfully save with question 2a2 unanswered  

image

ashleybot (approved):
@mlan225 Looks good. Good to bump versions up to `4.6.8` in this feature branch. Then after this feature is merged to main, tag that merge commit with v4.6.8 to trigger action to create nuget packages. 

For this one, hotfix commit COA.UDSv4 with new version of dependencies for UDS.Net.Forms and UDS.Net.Services. Be sure to test on your local before committing and pushing to remote.

PR #245, opened 2025-01-16
New Sorting Applied for A4

### LINKED ISSUE #236: A4 Medications Sorting
Medications in the A4 are incorrectly sorted. Currently they are using flex columns and the default ordering is across, we need them to sort by default downward. This is called _newspaper layout_.

In UDSv3 we handled this by finding the midpoint of the different medications lists then rendering in the appropriate column. Example:

```
    int defaultMedsMedian = (int)Math.Ceiling((double)defaultMeds.Count / 2);
```

```
    <div class="row">
        <div class="col">
            @foreach (var medicationSelected in defaultMeds.Take(defaultMedsMedian))
            {
                <partial name="_MedicationCheckbox" for="@medicationSelected" />
            }
        </div>
        <div class="col">
            @foreach (var medicationSelected in defaultMeds.Skip(defaultMedsMedian))
            {
                <partial name="_MedicationCheckbox" for="@medicationSelected" />
            }
        </div>
    </div>
```

ahmedhaly7777:
Fixes #236 

- [ ] Form A4 medicine names are laid out in a newspaper style, with names sorted vertically, starting in the left column and continuing in the right column.

### DIFF
--- a/src/UDS.Net.Forms.Tests/UDS.Net.Forms.Tests.csproj
+++ b/src/UDS.Net.Forms.Tests/UDS.Net.Forms.Tests.csproj
@@ -5,7 +5,7 @@
 		<ImplicitUsings>enable</ImplicitUsings>
 		<Nullable>enable</Nullable>
 		<IsPackable>false</IsPackable>
-		<ReleaseVersion>4.6.11</ReleaseVersion>
+		<ReleaseVersion>4.6.12</ReleaseVersion>
 	</PropertyGroup>
 	<ItemGroup>
 		<PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.11.1" />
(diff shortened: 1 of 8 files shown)


ashleybot (changes requested):
@ahmedhaly7777 Some of the code changes have broken the functionality of the form. Please test by checking a medication, saving, and then re-open the A4.

## To-do

- [ ] In this context, the `for` loop is safer than the `foreach` with a separate index property. More failure possibilities are introduced if that index doesn't get incremented. And the code is slightly less clean.
- [ ] The `Id` property mapping in some of the hidden fields has been deleted which is imperative.

ashleybot (changes requested):
@ahmedhaly7777 Just a couple small things to clean up.

Issue #246, opened 2025-01-16
A5D2 HEADINJUNC, HEADINJCON, HEADINJNUM, FIRSTTBI, LASTTBI Incorrect Required Validation


ashleybot:
Form cannot be finalized if TOBAC100 = 1 because required validation is incorrect.

Steps to replicate:
1. Choose TOBAC100 = 1
2. Answer all other required fields
3. Save as finalized
4. Form is returned with error messages for unrelated fields HEADINJUNC, HEADINJCON, HEADINJNUM, FIRSTTBI, LASTTBI

![Image](https://github.com/user-attachments/assets/86f17eac-166c-453c-a11e-5a80377a833c)

```
        [Display(Name = "After a head injury, what was the longest period of time that the participant was unconscious?")]
        [RegularExpression("^([0-4]|8|9)$", ErrorMessage = "Valid range is 0-4 or 8 or 9")]
        [RequiredIf(nameof(TOBAC100), "1", ErrorMessage = "Please specify.")]
        public int? HEADINJUNC { get; set; }
        [Display(Name = "After a head injury, what was the longest period that the participant was \"dazed or confused\" or unable to recall details of the injury?")]
        [RegularExpression("^([0-4]|8|9)$", ErrorMessage = "Valid range is 0-4 or 8 or 9")]
        [RequiredIf(nameof(TOBAC100), "1", ErrorMessage = "Please specify.")]
        public int? HEADINJCON { get; set; }
        [Display(Name = "Total number of head injuries in which the participant felt \"dazed or confused\", unable to recall details of the injury or experienced loss of consciousness?")]
        [RegularExpression("^([0-4]|9)$", ErrorMessage = "Valid range is 0-4 or 9")]
        [RequiredIf(nameof(TOBAC100), "1", ErrorMessage = "Please specify.")]
        public int? HEADINJNUM { get; set; }
        [Display(Name = "Age of first head injury that resulted in a period of feeling \"dazed or confused,\" being unable to recall details of the injury, or loss of consciousness")]
        [RegularExpression("^(\\d|[1-9]\\d|10\\d|110|999)$", ErrorMessage = "Valid range is 0-110 or 999")]
        [RequiredIf(nameof(TOBAC100), "1", ErrorMessage = "Please specify.")]
        public int? FIRSTTBI { get; set; }
[...]

PR #247, opened 2025-01-16
Adjust required attribute for sub question 4h

mlan225:
Resolves: #246 

## Adjust the sub question validation for section 4h (HEADINJURY) in form A5D2
Sub questions were incorrectly looking at the TOBAC100 value. Now, 4h sub questions will rely on the parent 4h question (HEADINJURY) to be 1 before requiring validation

- [x] if HEADINJURY (question 4h) is 1, then require all sub 4h sub questions to have a value
- [x] if HEADINJURY (question 4h) is 0 or 9, then 4h sub-questions are NOT required for completion

#### HEADINJURY == 1 (require input)
image

#### HEADINJURY == 0 || HEADINJURY == 9 (allow null)
image


ashleybot (approved):
@mlan225 Required validation is corrected and behavior functions as intended. Good to bump version to 4.6.10. Please update dependencies in UDS.COAv4 and deploy after business hours.

PR #258, opened 2025-01-24
Added labels

### LINKED ISSUE #237: A5D2 STROKESTAT radio button labels

ahmedhaly7777:
Fixes #237 

Added radio button labels to STROKSTAT question.


ashleybot (changes requested):
@ahmedhaly7777 I think valid HTML states that there can only be one `thead` per table. However, there can be unlimited `tbody`.

Can you change the `thead` to `tbody` and the `th` cells to `td`? And if you put a space after the forward slash then it will be able to flex more with different displays.

HTML reference => https://developer.mozilla.org/en-US/docs/Web/HTML/Element/tbody#multiple_bodies

ahmedhaly7777:
@ashleybot ready for review

ashleybot (approved):
@ahmedhaly7777 Good to merge to main!

PR #259, opened 2025-01-24
Map ZIP property to vm

### LINKED ISSUE #253: A1 ZIP Mapping to VM
Looking at the database it is clear the the ZIP is correctly being persisted. However, it is not being mapped properly back to the VM. So the input appears empty.

Oddvocado:
Fixes #253 

Although the Zip property was correctly persisted in the database, it was not being passed back to the view. The issue was resolved by mapping the property. Users should now be able to see the stored Zip value!


ashleybot (approved):
@Oddvocado Looks good! Please bump to version `4.6.14`

PR #260, opened 2025-01-24
A5d2 nomensage enable

### LINKED ISSUE #250: A5D2 NOMENSAGE = 99 should enable checkboxes in 7c
### Steps to replicate:

1. Create a new A5D2
2. Complete questions up until Section 7
3. Input 99 for `MENARCHE`
4. Input 99 for `NOMENSAGE`
5. Checkboxes in question 7c should be enabled, the are _incorrectly_ disabled and do not allow selection

Image

### Correct behavior (from step 4)

Checkboxes in question 7c should be enabled if `NOMENSAGE` is within range `10-70 or 99` and when this condition is true at least one of the checkboxes should be **required** upon finalization.

mlan225:
Resolves #250 

### Review checklist
- [x] Inputting a valid value of 10 - 70 or 99 in NOMENSAGE (7b) enables checkboxes in 7c
- [x] Inputting an invalid value that IS NOT 10 - 70 or 99 will disable the checkboxes in 7c
- [ ] Finalizing a packet with a value of ~99~88 in NOMENSAGE (7b) and no checkboxes selected in 7c will not produce an error message for no selected checkboxes
- [x] Finalizing a packet with a value of NOT 99 in NOMENSAGE (7b) and no checkboxes selected will produce an error message to select a checkbox

### 99 = enable checkboxes
image

### 88 = disable checkboxes
image

### valid range error for NOMENSAGE
image

### Validation error if no checkbox is selected when checkboxes are enabled (and NOT 99)
image

### DIFF
--- a/src/UDS.Net.Forms/Pages/UDS4/A5D2.cshtml.cs
+++ b/src/UDS.Net.Forms/Pages/UDS4/A5D2.cshtml.cs
@@ -1434,27 +1434,6 @@ public class A5D2Model : FormPageModel
         }
     };
 
-    public UIRangeToggle NOMENSAGEBehavior = new UIRangeToggle
-    {
-        Low = 10,
-        High = 70,
-        UIBehavior = new UIBehavior
-        {
-            PropertyAttributes = new List<UIPropertyAttributes>
-                {
-                    new UIEnableAttribute("A5D2.NOMENSNAT"),
-                    new UIEnableAttribute("A5D2.NOMENSHYST"),
-                    new UIEnableAttribute("A5D2.NOMENSSURG"),
-                    new UIEnableAttribute("A5D2.NOMENSCHEM"),
-                    new UIEnableAttribute("A5D2.NOMENSRAD"),
-                    new UIEnableAttribute("A5D2.NOMENSHORM"),
-                    new UIEnableAttribute("A5D2.NOMENSESTR"),
-                    new UIEnableAttribute("A5D2.NOMENSUNK"),
-                    new UIEnableAttribute("A5D2.NOMENSOTH"),
-                },
-        }
-    };
-
     public List<RadioListItem> BasicYesNoListItems { get; set; } = new List<RadioListItem>
     {
         new RadioListItem("No", "0"),
--- a/src/UDS.Net.Forms/Pages/UDS4/A5D2.cshtml.cs
+++ b/src/UDS.Net.Forms/Pages/UDS4/A5D2.cshtml.cs
@@ -1434,7 +1434,7 @@ public class A5D2Model : FormPageModel
         }
     };
 
-    public UIRangeToggle NOMESAGEBehavior = new UIRangeToggle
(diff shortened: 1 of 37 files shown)


src/UDS.Net.Forms/Pages/UDS4/A5D2.cshtml.cs
```diff
@@ -1434,27 +1434,6 @@ public A5D2Model(IVisitService visitService) : base(visitService, "A5D2")
...
-                    new UIEnableAttribute("A5D2.NOMENSCHEM"),
-                    new UIEnableAttribute("A5D2.NOMENSRAD"),
-                    new UIEnableAttribute("A5D2.NOMENSHORM"),
-                    new UIEnableAttribute("A5D2.NOMENSESTR"),
-                    new UIEnableAttribute("A5D2.NOMENSUNK"),
-                    new UIEnableAttribute("A5D2.NOMENSOTH"),
-                },
-        }
-    };
-
```
mlan225, line 1457: Replaced this behavior logic with javascript. After looking, I was unable to find a solution that would allow me to use the range behavior with a split range of 10 - 70 or 99. Instead of merging the 2 logics together and potentially risk them clashing logic, I decided to have the javascript handle the disabling here.

src/UDS.Net.Forms/wwwroot/js/A5D2.js
```diff
@@ -1,30 +1,54 @@
...
+            (radio) => radio.checked && radio.value == "1",
+        );
+
+        deprTreatRadios.forEach((radio) => {
+            radio.disabled = !(recentMajorDep || recentOtherDep);
+        });
+    }
+
+    function ToggleNOMENSAGECheckboxes() {
+        if (NOMENSAGEInput.value >= 10 && NOMENSAGEInput.value <= 70 | NOMENSAGEInput.value == 99) {
```
ashleybot, line 28: @mlan225 It is valid for the form to be finalized with `NOMENSAGE` = NULL (when MENARCHE is null). So, the case for `NOMENSAGE` without a value should be handled as well.

- [ ] Checkboxes should be disabled if NOMENSAGE does not have a value as well.

Screenshot 2025-01-28 at 1 32 45 PM

mlan225:
- [ ] Added validation to require checkbox selection when NOMENSAGE == 99

image

mlan225:
Adding a comment about it here @ashleybot , but currently NOMENSAGE has no requirement attributes and may be a worthy of an issue card. 

If i am reading the documentation correctly, it appears that NOMENSAGE has a "conditional missingness", which i interpret as "will sometimes be required under certain conditions". 
https://github.com/naccdata/uniform-data-set/blob/main/forms/uds/a5d2/form_a5d2_ivp_questions_and_vars.csv

image

ashleybot (approved):
@mlan225 NOMENSAGE changes look good! Good to bump to 5.0.2.

Issue #251, opened 2025-01-23
A5D2 MENARCHE UI behaviors


ashleybot:
### Steps to replicate:

1. Create a new A5D2 and complete questions up until Section 7
2. Input 88 for MENARCHE
3. NOMENSAGE is _incorrectly_ enabled

![Image](https://github.com/user-attachments/assets/6ee6a1b7-c81a-435e-8e4b-b753591cec79)

### Correct behavior

`NOMENSAGE` should only be enabled and required if `MENARCHE` is within range `5-25 or 99`

Documentation: https://github.com/naccdata/uniform-data-set/blob/main/forms/uds/a5d2/form_a5d2_ivp_questions_and_vars.csv

PR #261, opened 2025-01-27
Add js behavior for MENARCHE

mlan225:
resolves: #251

## Form A5D2 NOMENSAGE input should only be enabled and required if MENARCHE is within range 5-25 or 99
- [x] inputting a value of 5 - 25 or 99 will enable the NOMENSAGE field
- [x] inputting a non valid value will disable the NOMENSAGE input 
- [x] Finalizing a form with a MENARCHE input of 5 - 25 or 99 and a null value for NOMENSAGE will result in an error message

### Valid value, NOMENSAGE enabled
image

### Valid value, NOMENSAGE disabled
image

### valid value, NOMENSAGE not answered
image

### DIFF
--- a/src/UDS.Net.Forms/Pages/UDS4/A5D2.cshtml.cs
+++ b/src/UDS.Net.Forms/Pages/UDS4/A5D2.cshtml.cs
@@ -1419,21 +1419,6 @@ public class A5D2Model : FormPageModel
          },
     };
 
-    public UIRangeToggle MENARCHEBehavior = new UIRangeToggle
-    {
-        Low = 5,
-        High = 99,
-        UIBehavior = new UIBehavior
-        {
-            PropertyAttributes = new List<UIPropertyAttributes>
-                {
-                    new UIEnableAttribute("A5D2.NOMENSAGE"),
-                    new UIEnableAttribute("A5D2.HRT"),
-                    new UIEnableAttribute("A5D2.BCPILLS"),
-                },
-        }
-    };
-
     public UIRangeToggle NOMESAGEBehavior = new UIRangeToggle
     {
         Low = 10,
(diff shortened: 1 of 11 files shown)


mlan225:
The javascript for the A5D2 was reformatted for better readability when adding the additional methods and the controller behavior removed to handle complex range

src/UDS.Net.Forms/Pages/UDS4/A5D2.cshtml.cs
```diff
@@ -1419,21 +1419,6 @@ public A5D2Model(IVisitService visitService, IParticipationService participation
          },
     };
 
-    public UIRangeToggle MENARCHEBehavior = new UIRangeToggle
```
ashleybot, line 1422: @mlan225 Some functionality was lost when the ui behavior `MENARCHEBehavior` was removed. `HRT` and `BCPILLS` still need to have enabling and disabling behaviors.

- [ ] If MENARCHE is 5-25, 88, 99 then HRT and BCPILLS should be enabled
- [ ] If MENARCHE is null then HRT and BCPILLS should be disabled.

mlan225:
Added back in removed behavior for Menarche

#### Disable HRT and BCPILLS
When Menarche is an invalid range (NOT 5 - 25, 88, or 99), HRT and BCPILLS will disable. This includes NULL
image

#### Enable HRT and BCPILLS
When Menarche is a valid range, HRT and BCPILLS will be enabled
image

src/UDS.Net.Forms/Pages/UDS4/A5D2.cshtml.cs
mlan225: Good catch, i will add back in the behavior that was removed in an earlier commit, with a low / high range of 5 - 99.

ashleybot (approved):
@mlan225 Good to bump to version 6.0.2 and merge to main. Please make a hotfix commit to COA.UDSv4 and deploy to production outside of business hours.

**NOTE** I added a couple of small commits:

1. Required if validation for `HRT` and `BCPILLS` sub-questions
2. I also realized while reading the js that there was a small win to call `ToggleNOMENSAGECheckboxes()` and easily handle the nested scenario you had brought up. It doesn't handle every nesting scenario, but this was a clear, safe implementation for this one case.

PR #263, opened 2025-01-28
create and connect js controller file

### LINKED ISSUE #262: B6 GDS auto-calculation, and should allow 0
- [x] GDS score should autocalculate as user selecds each radio button option
- [x] If any question has an option of `Did not answer = 9` GDS score should be set to 88
- [x] GDS regular expression validation should be updated to allow `0`

Reference

- [NOGDS and GDS Rules](https://github.com/naccdata/uniform-data-set/blob/main/forms/uds/b6/form_b6_ivp_error_checks_p.csv)

mlan225:
Resolves #262 

## B6 Form GDS auto calculation (checklist updated) 
This PR focuses on the calculation of the fields

image

### Original Review Specs
- [x] GDS score should auto calculate as user selects each radio button option
- [ ] If any question has an option of Did not answer = 9 GDS score should be set to 88
- [x] GDS regular expression validation should be updated to allow 0
- [x] Form finalizes when all radios are input and NOGDS is false
- [x] Form gives an error when NOGDS is not checked and not all radio inputs are answered (unanswered radios display the error)
- [x] Form partial saves 

### Added Specs
- [x] GDS (question 16.) is a read-only field (from UDS3)
- [x] Unchecking the NOGDS checkbox will recalculate the GDS field with currently selected radio values
- [x] Radios selected while NOGDS is checked will result in the GDS total still being 88

### DIFF
--- a/src/UDS.Net.Forms/wwwroot/js/js_controllers/b6_controller.js
+++ b/src/UDS.Net.Forms/wwwroot/js/js_controllers/b6_controller.js
@@ -1,0 +1,17 @@
+import { Controller } from '@hotwired/stimulus';
+
+export default class extends Controller {
+
+    static targets = [
+        "radioTable",
+        "GDSInput"
+    ]
+
+    connect() {
+        //add on change to each radio input to run the calculation
+    }
+
+    CalculateGDS() {
+        //check each checked radio value and set GDS. If any radios are 9 then GDS is automatically 88
+    }
+}
--- a/src/UDS.Net.Forms/wwwroot/js/js_controllers/b6_controller.js
+++ b/src/UDS.Net.Forms/wwwroot/js/js_controllers/b6_controller.js
@@ -4,12 +4,14 @@ export default class extends Controller {
 
     static targets = [
         "radioTable",
-        "GDSInput"
+        "GDSInput",
+        "NOGDSInput"
     ]
 
-    connect() {
+    connect() { 
         this.CreateRadioButtonEvents()
         this.CalculateGDS()
+        this.NOGDSBehavior()
     }
 
     CalculateGDS() {
@@ -40,9 +42,19 @@ export default class extends Controller {
     }
 
     //GDS must be 88 if NOGDS checkbox is checked. Auto set for user on check
-    NOGDSBehavior(event) {
-        if (event.target.checked) {
+    NOGDSBehavior() {
+        if (this.NOGDSInputTarget.checked) {
             this.GDSInputTarget.value = 88
+            this.radioTableTarget.querySelectorAll('input[type="radio"]').forEach((radio) => {
+                radio.disabled = true
+            })
+        } else {
+            this.radioTableTarget.querySelectorAll('input[type="radio"]').forEach((radio) => {
+                radio.disabled = false
+            })
+
+            //recalculate GDS with enabled radios
+            this.CalculateGDS()
         }
     }
 }
--- a/src/UDS.Net.Forms/wwwroot/js/js_controllers/b6_controller.js
+++ b/src/UDS.Net.Forms/wwwroot/js/js_controllers/b6_controller.js
@@ -38,4 +38,11 @@ export default class extends Controller {
             })
         })
     }
+
+    //GDS must be 88 if NOGDS checkbox is checked. Auto set for user on check
+    NOGDSBehavior(event) {
+        if (event.target.checked) {
+            this.GDSInputTarget.value = 88
+        }
+    }
 }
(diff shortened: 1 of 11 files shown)


mlan225:
Spoke with Josh, bringing over functionality from UDS 3
- [ ] Disable radio buttons when NOGDS is checked

mlan225:
### Original Review Specs
- [ ] GDS score should auto calculate as user selects each radio button option
- [ ] If any question has an option of Did not answer = 9 GDS score should be set to 88
- [ ] GDS regular expression validation should be updated to allow 0

### Added Specs
- [ ] GDS (question 16.) is a read-only field
- [ ] When NOGDS is selected, radio inputs are disabled and GDS is set to 88
- [ ] Unchecking the NOGDS checkbox to reenable the radio fields will recalculate the GDS field with currently selected radio values

src/UDS.Net.Forms/wwwroot/js/js_controllers/b6_controller.js
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    CalculateGDS() {
+        //If "Did not answer" radio detected, stop calculation loop
+        let noCompleteFlag = false
+
+        //reset values for recalculation
+        this.GDSInputTarget.value = ""
+
+        this.radioTableTarget.querySelectorAll('input[type="radio"]:checked').forEach((radio) => {
```
mlan225, line 22: Wanted to use a variable to control checked and unchecked radio button DOM queries, however, setting the variables once results in the elements load state saved. So using another query selector on the stimulus target insures the elements are reselected. 

A little more repeated, but I'm not as worried to have DOM selectors reused in the different methods in this case. Also makes it more understandable when reading each method.

src/UDS.Net.Forms/wwwroot/js/js_controllers/b6_controller.js
```diff
@@ -0,0 +1,48 @@
+import { Controller } from '@hotwired/stimulus';
```
mlan225, line 1: Using raw javascript in controller, Cody has been wanting to stick with raw JS so I didn't want to toss in Jquery so our controllers are consistent.

src/UDS.Net.Forms/wwwroot/js/js_controllers/b6_controller.js
ashleybot: @mlan225 Would it be possible to use ui-behaviors here rather than including new js?

mlan225: Yeah, I can have the UI behavior manage the checkbox disabling of the radio buttons and then have the values calculated with JS

mlan225:
Modified the logic to remove disabling the radio buttons. Spoke with Ash about the form logic and we agreed to revisit the disable validation in a later card after inspecting the form guidelines further

ashleybot (changes requested):
@mlan225 Great progress on this (and thanks for the super clear description in the PR - it helped a ton!)

I think that this interpretation is slightly off

> If any question has an option of Did not answer = 9 GDS score should be set to 88

I think it should be

**If 4 or more questions have an option of Did not answer = 9 GDS score should be set to 88**

Here's NACC's documentation

> If at least 12 of (SATIS, DROPACT, EMPTY, BORED, SPIRITS, AFRAID, HAPPY, HELPLESS, STAYHOME, MEMPROB, WONDRFUL, WRTHLESS, ENERGY, HOPELESS, BETTER) in (0,1), then GDS must equal sum of all subquestions in (0,1)

I think that means two things:

-  if there are 12 or more questions answered (`0 or 1`), then the sum should be the total of all subquestions (like the regular calculation).
- if there are less than 12 questions answered, meaning 4 or more are answered `9`, then that is when the `GDS` score should be 88 and `NOGDS` = true

mlan225:
### Adjust validation rules and JS functionality
- [ ] When selecting NGDS, the total will be set to 88
- [ ] When selecting 9 (did not answer) for any of the questions, total score javascript will no longer change to 88 and instead, will continue to be the sum of the selected questions
- [ ] The javascript has been adjusted to only set total score to 88 when NGDS has been selected
- [ ] The validation method on the model `NOGDSCheckedValidation()` now checks for less than 12 questions to be 1 or 0 before requiring NOGDS to be selected and GDS to be 88
- [ ] The validation method on the model `NOGDSCheckedValidation()` now returns a validation message if 12 or more questions are answered 1 or 0 AND NOGDS is selected and GDS is 88


### Extra validation
- [ ] method `GDSumAndNOGDSValidation()` on the model will make sure that if NOGDS is selected then GDS must be 88. Our javascript will make sure this will never happen, as it sets total to 88 when NOGDS is checked, however, I felt I needed to double down and include this ruling on the model so it can state all validation rules regardless of what the javascript is doing.


image

ashleybot (approved):
@mlan225 Nice additional validation! This helped in my understand the form a lot more. 

Good to bump to version 6.0.3 and merge to main. Please make a hotfix commit to COA.UDSv4 and deploy outside of business hours.

PR #265, opened 2025-01-31
Implements "Save and go next" button

ashleybot:
- Uses service method to get the next form kind and ordering from service layer (so it is customizable for consuming apps)
- Adds "save and go next" button to form footer

- [x] Save and go next button only renders **if** there is a next form

- Provides functionality in base FormPageModel class via property on POST

- [x] Functionality is backward compatible and property not required so this will not introduce a breaking change to consuming apps


Oddvocado:
@ashleybot looks good and a cool implementation! I was able to save the form and continue to the next form when available. As expected, the D1b form does not provide the option to save and continue.

Oddvocado (approved):

Issue #282, opened 2025-02-19
D1a PREDOMSYN = 1 and required Section 3 syndrome present bug


ashleybot:
Currently the D1a required a syndrome in Section 3 to be indicated as present if PREDOMSYN = 1. When PREDOMSYN = 1 then a selection in Section 3 should not be required.

Steps to replicate:

1. Complete the form with all the required fields with a minimum of PREDOMSYN = 1
2. Leave Section 3 blank
3. Attempt to finalize the form

![Image](https://github.com/user-attachments/assets/71d9c7c6-e70f-4420-abe3-d85283b739bd)

![Image](https://github.com/user-attachments/assets/8cffc42e-4f93-4ca1-a3c3-46058331e40e)

PR #289, opened 2025-02-26
update the required attribute for PresentSyndromeIndicated

mlan225:
Resolves: #282 

## PREDOMSYN requirement
This PR allows section 3 to not be answered if question 8 (PREDOMSYN) is 1. This, in turn, will make section 3 values required when question 8 is 0

- [x] If PREDOMSYN = 1 then a selection in section 3 should not be required
- [x] a selection in section 3 is required when PREDOMSYN = 0

## This PR has undergone some additional scope changes 
In light of some rule clarification, we adjusted some additional validation on this form **[Copied from comment below]**

#### Question 4
image
- [x] if all checkboxes are checked question 4b (MCI) must be equal to 1
- [x] if not all check boxes are checked question 4b (MCI) must be equal to 0
- [x] if no checkboxes are selected MCI must be 0

##### Question 5
image
- [x] if any of the checkboxes from question 4 are checked **With the exception of JUST the third checbox** OR if any of the checkboxes from question 5 are checked, then question 5b (IMPNOMCI) should be 1 - Yes
- [x] if **ONLY** the third checkbox in question 4 is checked and no options in question 5 are checked, then question 5b (IMPNOMCI) should be 0 - No
- [x] When PREDOMSYN (question 8) is 1 at least 1 sub question from 8a - 8l should be marked as "present" AND at least one checkbox from question 9 should be checked
- [x] When PREDOMSYN (question 8) is 0 at least one question from 10 - 32 should be selected as present

### DIFF
--- a/src/UDS.Net.Forms/Pages/UDS4/D1a.cshtml
+++ b/src/UDS.Net.Forms/Pages/UDS4/D1a.cshtml
@@ -193,7 +193,6 @@
             <span class="counter"></span>Check all applicable criteria for cognitively impaired, not MCI/dementia in Q5, using any relevant data. Any conditions
             contributing to impairment (e.g., substance abuse or medications) should be identified in Section 3.
         </label><br />
-        @* <span class="mt-2 text-sm text-red-600" asp-validation-for="D1a.CognitivelyImpairedIndicated"></span> *@
     </div>
     <div class="mt-4 sm:col-span-2 sm:mt-0">
         <fieldset class="subsubcounterreset">
--- a/src/UDS.Net.Forms/Pages/UDS4/D1a.cshtml
+++ b/src/UDS.Net.Forms/Pages/UDS4/D1a.cshtml
@@ -122,7 +122,6 @@
     <div class="ml-2 sm:grid sm:grid-cols-3 sm:items-start sm:gap-4 sm:py-6">
         <div>
             <label class="text-base font-semibold text-gray-900"><span class="counter subcounterreset"></span>Check all criteria that apply in Q4.</label><br />
-            <span class="mt-2 text-sm text-red-600" asp-validation-for="D1a.MCIClinicalCriteriaIndicated"></span>
         </div>
         <div class="mt-4 sm:col-span-2 sm:mt-0">
             <fieldset class="subsubcounterreset">
@@ -580,6 +579,7 @@
     </div>
 </div>
 <div class="sm:grid sm:grid-cols-1 sm:items-start sm:gap-4 sm:py-6">
+    <span class="mt-2 text-sm text-red-600" asp-validation-for="D1a.PREDOMSYNSyndromePresent"></span>
     <table class="ml-2 table-auto min-w-full divide-y divide-gray-300">
         <thead>
             <tr class="odd:bg-white even:bg-gray-50 border-b">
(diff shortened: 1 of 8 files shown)


mlan225:
## Notes to self 

After looking into production data, it appears that PREDOMSYN has never had a value. All completed forms on our UDS4 were completed without providing a value for PREDOMSYN, which doesn't seem to be allowed in our current build of the application. I wonder if a recent change adjusted the validation for this? 

### UDS4 NACC Rules
The UDS4 rules state that if no symptoms are present, then mark 0.

image


### Our UDS validation requirements
D1a PREDOMSYN requirements

mlan225:
## D1a NACC documentation and SBCoA current implementation flow

D1a flow

mlan225:
Identified a bug in the form. **Low priority**

- FORM END disable state remains on question 6a if quest 2b is answered on question 2 change until form reload

image
image
image


The javascript surrounding these 2 options is causing it to be impossible to continue the form on question 6a if you attempt to select these options, but then change question 2 (NORMCOG) to option 0 - skip to question 3 until the form is reloaded.

### Steps to reproduce
1. select option 1 for question 2
2. select option 1 for question 2a
3. select any of the options for 2b to END FORM HERE
4. change the value for question 2 to 0 and continue form
5. checkboxes in 6a will remain disabled until reload

mlan225:
### Updates to the PR
#### Question 4
- [ ] if all checkboxes are checked question 4b (MCI) must be equal to 1 or an error message will return on finalize
- [ ] if not all check boxes are checked question 4b (MCI) must be equal to 0 or an error message will return on finalize

##### Question 5
- [ ] if any of the checkboxes from question 4 are checked **With the exception of JUST the third checbox** OR if any of the checkboxes from question 5 are checked, then question 5b (IMPNOMCI) should be 1 - Yes or an error message will return on finalize
- [ ] if **ONLY** the third checkbox in question 4 is checked and no options in question 5 are checked, then question 5b (IMPNOMCI) should be 0 - No or an error message will return on finalize 
- [ ] When PREDOMSYN (question 8) is 1 at least 1 sub question from 8a - 8l should be marked as "present" AND at least one checkbox from question 9 should be checked or an error message will return 
- [ ] When PREDOMSYN (question 8) is 0 at least one question from 10 - 32 should be selected as present or an error message will return on finalize

mlan225:
Updates to PR
- [ ] require MSI to be 0 when question 4 has 0 - 2 checkboxes selected

src/UDS.Net.Forms/Pages/UDS4/D1a.cshtml
```diff
@@ -193,7 +193,7 @@
             <span class="counter"></span>Check all applicable criteria for cognitively impaired, not MCI/dementia in Q5, using any relevant data. Any conditions
             contributing to impairment (e.g., substance abuse or medications) should be identified in Section 3.
         </label><br />
-        <span class="mt-2 text-sm text-red-600" asp-validation-for="D1a.CognitivelyImpairedIndicated"></span>
+        @* <span class="mt-2 text-sm text-red-600" asp-validation-for="D1a.CognitivelyImpairedIndicated"></span> *@
```
ashleybot, line 196: @mlan225 Can this be removed rather than commented out? It looks like CognitivelyImpairedIndicated was replaced with IMPNOMCICriteriaValidation, correct? This comment in the view could cause confusion if left with no explanation.

ashleybot (commented):
@mlan225 Looks good! This is nearly ready. Just a small line commented out in the view that could cause confusion in the future.

Can you also upload your flow diagram to the wiki? Here =>https://github.com/UK-SBCoA/uniform-data-set-dotnet-web/wiki/D1a-Clinical-Diagnosis

src/UDS.Net.Forms/Pages/UDS4/D1a.cshtml
mlan225: Yeah, that simple check for making sure a checkbox was selected was replaced with the more complex validation. I went ahead and removed it

also, yeah! I can place it in there

PR #293, opened 2025-03-04
Added validation

### LINKED ISSUE #233: Add custom validation for MOCABTOTS

ahmedhaly7777:
Fixes #233 

Added validation for `MOCBTOTS`

The condition is that questions `1e–1k` and `1n–1s` must be between 95 and 98, and `MOCBTOTS != 88`. 
An error message will be shown: "`If 1e–1k or 1n–1s were not administered, then MOCBTOTS must be 88`".


Oddvocado (changes requested):
@ahmedhaly7777 The implementation of the ValidationResult class works great! I encountered the expected validation errors when the following NACC requirement was not meant:

> 	If any of MOCADIGI, MOCALETT, MOCASER7, MOCAREPE, MOCAFLUE, MOCAABST, MOCARECN, MOCAORDT, MOCAORMO, MOCAORYR, MOCAORDY, MOCAORPL, MOCAORCT are 95-98 then MOCBTOTS must be 88

However, I was able to save the form as "Finalized" without providing a value for MOCBTOTS. This question will need to have a annotation placed above the property to prevent this from happening. I believe there is a  [RequiredIfTelephoneVisit] attribute that can be applied.

Issue #314, opened 2025-03-11
Remove TRSPLONGER from A1a


ashleybot:
NACC form no longer has TRSPLONG in A1a

![Image](https://github.com/user-attachments/assets/e5d721a6-14b9-45c4-abc6-fb8e9fc6e039)

https://files.alz.washington.edu/documentation/uds4-ivp-a1a.pdf

Issue #315, opened 2025-03-11
Remove COMPUSA and FAMCOMP from A1a


ashleybot:
- Remove COMPUSA
- Remove FAMCOMP
- Include option to select 88 for COMPCOMM

https://files.alz.washington.edu/documentation/uds4-ivp-a1a.pdf

![Image](https://github.com/user-attachments/assets/e530a358-0f9d-4314-836d-dac27e2a1524)

PR #322, opened 2025-03-11
 Remove TRSPLONGER from A1a

Oddvocado:
Fix: #314 

This PR removes the TRSPLONGER  property so that the A1a form is up to date with the NACC spec

### DIFF
--- a/src/UDS.Net.Forms/Extensions/DomainToViewModelMapper.cs
+++ b/src/UDS.Net.Forms/Extensions/DomainToViewModelMapper.cs
@@ -514,7 +514,6 @@ namespace UDS.Net.Forms.Extensions
                 TRSPACCESS = fields.TRSPACCESS,
                 TRANSPROB = fields.TRANSPROB,
                 TRANSWORRY = fields.TRANSWORRY,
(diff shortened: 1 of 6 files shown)


Oddvocado:
Adding issue #315 to this PR as well since it involves the same form

src/UDS.Net.Forms/Pages/UDS4/A1a.cshtml.cs
```diff
@@ -84,6 +84,12 @@ public A1aModel(IVisitService visitService, IParticipationService participationS
             new RadioListItem("1", "1"),
         };
 
+        public List<RadioListItem> PreferNotToAnswer { get; } = new List<RadioListItem>
```
ashleybot, line 87: @Oddvocado This option needs to be added to `LadderItems` not an entirely new list.

Oddvocado:
@ashleybot thanks for making that change! I also edited the display annotation of COMPCOMM so that it informs the user that a response of 88 indicates that the participant did not wish to answer the question.

ashleybot (approved):
@Oddvocado Looks great! Please merge to main, no need to update version, we'll do that with the last card.

PR #325, opened 2025-03-11
Removed

### LINKED ISSUE #317: Remove NEWINF from A2

ahmedhaly7777:
Fixes #317


src/UDS.Net.Forms/Models/UDS4/A2.cs
```diff
@@ -65,8 +58,7 @@ public override IEnumerable<ValidationResult> Validate(ValidationContext validat
                     {
                         if (visit.PACKET == PacketKind.F)
```
ashleybot, line 59: @ahmedhaly7777 This whole conditional should be removed

```
                        if (visit.PACKET == PacketKind.F)
                        {
                            yield return new ValidationResult("Response required");
                        }
```

If left, it will not allow finalization on any follow-up packet, and adding ValidationResult without specifying a property will have unintended side-effects.

ashleybot (changes requested):
@ahmedhaly7777 Just a little more cleanup in the custom validation.

ashleybot (approved):
@ahmedhaly7777 Good to merge to main, no need to adjust version numbers.

PR #331, opened 2025-03-12
FTLDIF Removed 

### LINKED ISSUE #298: D1b fix label numbering
Question 14 sub-questions are using a css numbering scheme that is one level too deep. Adjust so that the sub-questions are 14a through 14e.

## Our app on the left, NACC pdf on the right

Image

### LINKED ISSUE #321: Remove FTLDIF from D1b

ahmedhaly7777:
Fixes #321 
Fixes #298


ashleybot (changes requested):
@ahmedhaly7777 Use the solution-side search feature to search for the property. Needs to be removed from DomainToViewModelMapper. Also, does not build. Always make sure your solution builds before submitting PR to be reviewed.

ashleybot (changes requested):
@ahmedhaly7777 Much better. This would be a good time to address #298 along with this PR. Question 14 has a couple extra things to adjust in the UI to match the specification (mainly the label numbering). 

- [ ] Adjust span css classes for subcounter and subsubcounter

There are some intricacies with css nested and resetting counters, so let me know if you hit a roadblock.

Screenshot 2025-03-12 at 1 04 10 PM

ashleybot (approved):
@ahmedhaly7777 Looks great! Thanks for shoring up the numbering for question 14 sub-questions. Good to merge to main, do not need to bump version.

Issue #353, opened 2025-03-19
Adjust headers to be lowercase


ashleybot:
Found in NACC documentation that headers should be lowercase.

![Image](https://github.com/user-attachments/assets/84e73906-f001-40ed-bd59-e2e5f4204bb4)

### Reference
https://docs.naccdata.org/edc/uploads-to-nacc/file-submission

mlan225:
https://joshclose.github.io/CsvHelper/getting-started/#writing-a-csv-file

PR #355, opened 2025-03-19
Adjust export to make lower case headers

mlan225:
Resolves: #353 

## Adjust export to make lower case headers
- [ ] Export columns in csv are all lower case
- [ ] Export matches column count with NACC sample file

### DIFF
--- a/src/UDS.Net.Forms/Pages/PacketSubmissions/Export.cshtml.cs
+++ b/src/UDS.Net.Forms/Pages/PacketSubmissions/Export.cshtml.cs
@@ -390,7 +390,7 @@ namespace UDS.Net.Forms.Pages.PacketSubmissions
             {
                 if (property != null)
                 {
-                    if (property.PropertyType.IsPrimitive || property.PropertyType == typeof(int?) || property.PropertyType == typeof(bool?) || property.PropertyType == typeof(decimal?) || property.PropertyType == typeof(double?) || property.PropertyType == typeof(string))
+                    if (property.PropertyType.IsPrimitive || property.PropertyType == typeof(int?) || property.PropertyType == typeof(bool?) || property.PropertyType == typeof(decimal) || property.PropertyType == typeof(decimal?) || property.PropertyType == typeof(double?) || property.PropertyType == typeof(string))
                     {
                         csv.WriteField(property.Name.ToLower());
                     }
--- a/src/UDS.Net.Forms/Pages/PacketSubmissions/Export.cshtml.cs
+++ b/src/UDS.Net.Forms/Pages/PacketSubmissions/Export.cshtml.cs
@@ -387,10 +387,12 @@ namespace UDS.Net.Forms.Pages.PacketSubmissions
         {
             foreach (var property in typeof(T).GetProperties())
             {
+                if (property.PropertyType.IsPrimitive || property.PropertyType == typeof(int?) || property.PropertyType == typeof(bool?) || property.PropertyType == typeof(decimal?) || property.PropertyType == typeof(double?) || property.PropertyType == typeof(string))
+                {
+                    string propertyName = property.Name.ToLower();
 
-                string propertyName = property.Name.ToLower();
-
-                csv.WriteField(propertyName);
+                    csv.WriteField(propertyName);
+                }
             }
         }
     }
--- a/src/UDS.Net.Forms/Pages/PacketSubmissions/Export.cshtml.cs
+++ b/src/UDS.Net.Forms/Pages/PacketSubmissions/Export.cshtml.cs
@@ -82,22 +82,22 @@ namespace UDS.Net.Forms.Pages.PacketSubmissions
                     if (a1 != null)
                     {
                         csv.WriteHeader<A1Record>();
-                        WriteFormFields(new A1FormFields(), csv);
+                        csv.WriteHeaderLowercase<A1FormFields>();
                     }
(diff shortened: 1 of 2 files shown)


mlan225:
After removing the code that compares all columns on as lower case, the file matches the NACC export

first file = uds4 export
second file = NACC sample export

image

mlan225:
Going to see if i can refactor the `WriteFormFields` method if statement. 

Using a method to go through properties of each formfield file to replicate the `WriteHeader` functionality. Instead of attempting to override the csv WriteHeader method, I used my own method to set column names as lower case in the formFields files. Record files are already using lower case attributes and are not modified. 

Noticed that the WriteHeader method was only grabbing primitive types, In my method I'm checking for primitive type and primitive types that have been made nullable (making them no longer primitive). There may be a better way to write this.

mlan225:
Will push an update to the COA csv checker to remove the lower case code after changing the export columns to lower case

mlan225:
geGetProperties seems to return assembly order from .net 7 onward, so that's convenient 

> In .NET 6 and earlier versions, the [GetProperties](https://learn.microsoft.com/en-us/dotnet/api/system.type.getproperties view=net-9.0) method does not return properties in a particular order, such as alphabetical or declaration order. Your code must not depend on the order in which properties are returned, because that order varies. However, starting with .NET 7, the ordering is deterministic based upon the metadata ordering in the assembly.

src/UDS.Net.Forms/Pages/PacketSubmissions/Export.cshtml.cs
```diff
@@ -382,6 +382,21 @@ public async Task<IActionResult> OnGetAsync(int packetId)
...
+        private void WriteFormFields<T>(T formFields, CsvWriter csv)
+        {
+            foreach (var property in typeof(T).GetProperties())
+            {
+                if (property != null)
+                {
+                    if (property.PropertyType.IsPrimitive || property.PropertyType == typeof(int?) || property.PropertyType == typeof(bool?) || property.PropertyType == typeof(decimal) || property.PropertyType == typeof(decimal?) || property.PropertyType == typeof(double?) || property.PropertyType == typeof(string))
+                    {
+                        csv.WriteField(property.Name.ToLower());
+                    }
```
mlan225, line 396: Attempted to find a more dynamic way to handle this check, but this seems to be the solution. The `WriteHeader` method of the CSVHelper seems to only target primitive types as well as primitive types that are nullable (making them no longer primitive). 

At the moment, we conveniently are only sending these simple types to NACC

Added type non-nullable `decimal` to the checks. While type `decimal` is not currently being used in formfields, it is also NOT a primitive type and won't be caught by `property.PropertyType.IsPrimitive` in case we use it in the future.

mlan225, line 396: Not running an error check here at the end because if for some reason the field fails on the `ToLower` method, we want the program to crash and the error be addressed.

src/UDS.Net.Forms/Pages/PacketSubmissions/Export.cshtml.cs
```diff
@@ -382,6 +382,21 @@ public async Task<IActionResult> OnGetAsync(int packetId)
             }
             return null;
         }
+
+        // Custom write method for formfields to convert property names to lower case
+        private void WriteFormFields<T>(T formFields, CsvWriter csv)
```
ashleybot, line 387: @mlan225 I like your thinking for extensibility. Take a look at the source code for some examples and let's change a couple things:

- [ ] Since we are only going to use this functionality for header rows only then let's name this accordingly, and we can follow similar naming to the library with WriteHeader<T> and make it an extension method so we use it in the same way
- [ ] Create a new `static` class in the Extensions directory CsvHelperExtensions.cs
- [ ] Move your method there and make it `public` `static` and use `this` modifier on the parameter to make it an extension method

```
public static void WriteHeaderLowercase<T>(this CsvWriter csv)
{
...
// your code here, check out the source code for additional considerations
}
```

This is an example of the way you'll use it:
```
 csv.WriteHeaderLowercase<A1FormFields>();
```
Instantiating the form fields is unnecessary because you are only using reflection to inspect the type.

Overall, this is a small change that will benefit us to be more consistent and also in the future if we need to make more adjustments to the way the header is rendered. 

### References

Source code for CsvHelper WriteHeader => https://github.com/JoshClose/CsvHelper/blob/852bd465e0d460eb60013abdc07fff8f68f68a59/src/CsvHelper/CsvWriter.cs#L238

src/UDS.Net.Forms/Pages/PacketSubmissions/Export.cshtml.cs
ashleybot: @mlan225 I think it's fine to keep this, but remember that this effort is only for the header row and not for the data rows. There should only be string variable names in the header and no other types.

ashleybot: @mlan225 In this case, I don't think we would want the app to crash, we'd want the export and to be able to open and see the messed up file.

ashleybot (changes requested):
@mlan225 Some small adjustments to make the method an extension method.

ashleybot (approved):
@mlan225 Nice work!

PR #368, opened 2025-03-28
In progress adjustments

### LINKED ISSUE #367: Export Converter should export null instead of 0 for bool

mlan225:
Resolves #367


mlan225:
Bool conversion is a little more tricky than I had originally thought.

We identified that bools are sometimes converted to integer values in the form fields. This was due to needing an integer value for the checkboxes to render in the views. 

There IS  a way to go about using a nullable boolean in the form instead, it just takes a little bit of conversion refactoring to make sure the proper values are being passed to the models

I've set this up with `ETHGERMAN` and may continue with this path input by input.

mlan225:
There is also a quicker approach available to override the default converter instead of just the boolean converter so that it targets each input. This would make sure to also target formfields that are set for an `int?` type, as it currently does not.

From there we could create a manual list of properties by name that go from bool to int on the formfields to handle those, while supplying code in the new method to handle each property type.

PR #374, opened 2025-04-01
Update D1b.cs

### LINKED ISSUE #344: D1b New PET/SPECT enhanced validation
Add a new NotMapped property for enhanced validation for the PET/SPECT section

### Specification

If AMYLPET and TAUPET are NOT NULL then at least one must not be 8.

- [x] Add the NotMapped property to the D1b view model
- [x] Display the error message in a proximal location to the "If used in diagnosis, indicated the results" table
- [x] Extensively test all failure and success states including correct failures

### Failure state
Since there are only two properties to consider with this specification, there is only one failure state.

Image

### Reference

- [NACC specifications line 53](https://github.com/naccdata/uniform-data-set/blob/main/forms/uds/d1b/form_d1b_ivp_error_checks_mc.csv)

ahmedhaly7777:
Fixes #344


ashleybot (approved):
@ahmedhaly7777 We don't have the different ways to implement validation or any conceptual documentation, but generally when you make a `NotMapped` property to handle validation then you should display it with the `asp-validation-for` helper. This simplifies and clarifies how validation is performed. 

Adding code to the overridden `Validate` method should only be used for more advanced scenarios when you can't accomplish what you want with the `NotMapped` property.

So, use one or the other implementation for validation, but never both.

Please see my commit for more details. And let's talk through things if you'd like more details.

PR #381, opened 2025-04-07
 Form Mode and Reasons

### LINKED ISSUE #362: Form Mode and Reasons
In the form footer, the form's mode can be selected. [Form modes](https://github.com/UK-SBCoA/uniform-data-set-dotnet-web/blob/main/src/UDS.Net.Services/Enums/FormMode.cs):

1. In-person - all forms can be conducted in-person
2. Remote - only some forms can be conducted remotely, if it isn't an option then this is disabled
3. Not completed - only some forms are optional, in required forms this is disabled

When remote is selected and a reason is chosen, if the user changes back to in-person mode they have no way of  clearing reason or modality. So, add some js to accomplish this for them.

## Select remote mode
Image

## Select a remote reason
Image

## Select in-person mode
You cannot clear mode because the empty option is disabled. So, let's clear it for the user if they change the mode to in-person
Image

- [ ] When mode is changed to in-person or not completed, clear remote reason and remote modality if they have a value
- [ ] When the mode is changed to in-person or remote, clear the not completed reason if it has a value

## Reference
- Remote [Reason](https://github.com/UK-SBCoA/uniform-data-set-dotnet-web/blob/main/src/UDS.Net.Services/Enums/RemoteReasonCode.cs) and [Modality](https://github.com/UK-SBCoA/uniform-data-set-dotnet-web/blob/main/src/UDS.Net.Services/Enums/RemoteModality.cs)
- Not completed [reason codes](https://github.com/UK-SBCoA/uniform-data-set-dotnet-web/blob/main/src/UDS.Net.Services/Enums/NotIncludedReasonCode.cs)

Oddvocado:
Fixes #362 

This PR adds JavaScript to dynamically reset dropdown values in the form footer partial based on the currently selected mode. For example, if the Mode is changed from "Remote" to "InPerson", any selections in the "Remote Reason" and "Remote Modality" dropdowns will be reset to their default values.

Additionally, I’ve implemented logic to disable certain dropdown options depending on the selected mode. If "InPerson" is selected, all "Remote Reason", "Remote Modality", and "Not Completed Reason" options will be disabled.

This disabling behavior is persistent after form submission — if a form is submitted with "Remote" selected, the "Not Completed Reason" options should remain disabled when the page is revisited.

Screenshot 2025-04-09 154933


Oddvocado:
@ashleybot Added a stimulus controller and removed the old js file. Ready for another review!

ashleybot (changes requested):
@Oddvocado So, I'm seeing now this implementation actually needs to be specific and we won't be able to make it generic.

Oddvocado:
@ashleybot refactored the code to utilize stimulus targets. Ready for another review!

ashleybot (approved):
@Oddvocado This is fantastic! 🏆  Good to increase patch version and merge to main.

Issue #383, opened 2025-04-09
PTID 1944 submission errors


mlan225:
# Fix errors with submission PTID 1944 

### Date incorrectly formatted
- [ ] Switch dates to MM-dd-yyyy from dd-MM-yyyy

mlan225:
Moving to blocked, as one of the errors here relate to a card @ahmedhaly7777 is working on

Issue #370, opened 2025-03-31
C2 VERBALTEST UI Behaviors


ashleybot:
UI Behaviors need to be adjusted when VERBALTEST = 1 and TRIAL 1 95-97

> if q12. verbaltest (verbal learning test) is 1 then q13a. reydrec (delayed recall) cannot be blank

So, Sections 12 and 13 should **not** be both disabled.

Trial 1 should only disable the remaining section 12, and REYDREC should control the remainder on section 13.

![Image](https://github.com/user-attachments/assets/1609c87a-8e4d-4a39-a4f5-39fa2815dfa2)

### Reference

Error returned from NACC submission
![Image](https://github.com/user-attachments/assets/63d49f8e-7d7e-47ef-9ee4-2845c24e201a)

ashleybot:
Question was forwarded to NACC research team. Awaiting response.

ashleybot:
We overwrote REYDREC with 97 and attempted the submission last Friday and it worked. Still awaiting NACC's response.

PR #384, opened 2025-04-09
PTID error fixes

mlan225:
Fixes: #383 
Fixes: #370

# Fix errors with submission

### form date incorrectly formatted for record export (all records)
- [x] Switch dates to MM-dd-yyyy from dd-MM-yyyy for all records

### C2 form errors
- [x] If REY1REC is 95 - 98, then REYDREC must have a value of 95 - 98 
  - This is additional validation to REYDREC being required if REY1REC = 0 - 15
- [x] Entering 95 - 98 for REY1REC no longer disables REYDREC
- [x] Update instruction text under question 12a (REY1REC) in-person/video modalities to: "If test was not completed, enter reason code, 95-98. Complete QUESTION 13a and SKIP TO QUESTION 16a."
- [x] Remove reference to value 88 from validation for REYDREC

### DIFF
--- a/src/UDS.Net.Forms/wwwroot/js/C2.js
+++ b/src/UDS.Net.Forms/wwwroot/js/C2.js
@@ -36,3 +36,16 @@ document.addEventListener('turbo:frame-load', function (event) {
         }
     }
 });
+
+//Set REYDREC to REY1REC when REY1REC is 95 - 98
+let REY1RECInput = document.getElementById("C2_REY1REC")
+let REYDRECInput = document.getElementById("C2_REYDREC")
+
+REY1RECInput.addEventListener('change', () => {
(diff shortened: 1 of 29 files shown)


src/UDS.Net.Forms/wwwroot/js/C2.js
```diff
@@ -36,3 +36,16 @@ document.addEventListener('turbo:frame-load', function (event) {
         }
     }
 });
+
+//Set REYDREC to REY1REC when REY1REC is 95 - 98
```
mlan225, line 40: Stimulus did not seems to be cooperating, so I made the feature in the existing c2.js file

ashleybot (approved):
@mlan225 Looked through the changes. Good to merge!

Some feedback to keep in mind (does **not** need to be changed now):

- If you have a common format string like that date formatter, you can set it to a public const and use that variable. That way if the spec changes, or you need to fix a bug, then it only requires a change in one place.

Issue #394, opened 2025-04-18
Form C2 MOCATOTS enhanced validation


mlan225:
Submissions to NACC revealed further validation necessary

> If all subquestions are completed, MOCATOTS is the total of (MOCATRAI, MOCACUBE, MOCACLOC, MOCACLON, MOCACLOH, MOCANAMI, MOCADIGI, MOCALETT, MOCASER7, MOCAREPE, MOCAFLUE, MOCAABST, MOCARECN, MOCAORDT, MOCAORMO, MOCAORYR, MOCAORDY, MOCAORPL, MOCAORCT)	

## MOCATOTS sub question validation rules
These rules apply to the (in-person & video modalities)

#### Target questions:

1. MOCATRAI, 
2. MOCACUBE, 
3. MOCACLOC, 
4. MOCACLON, 
5. MOCACLOH, 
6. MOCANAMI, 
7. MOCADIGI, 
8. MOCALETT, 
9. MOCASER7, 
10. MOCAREPE, 
11. MOCAFLUE, 
12. MOCAABST, 
13. MOCARECN, 
14. MOCAORDT, 
15. MOCAORMO, 
16. MOCAORYR, 
17. MOCAORDY, 
18. MOCAORPL, 
19. MOCAORCT

- [ ] Make the value of MOCATOTS required to be equal to sum of the target questions if those questions are completed
- [ ] Do not require MOCATOTS to the sum when the target questions are not ALL completed
> While all sub questions are required, having a value of 95 - 98 for any target questions will consider the sub questions "not all completed"

PR #396, opened 2025-04-23
C2 MOCATOTS sum validation

mlan225:
Resolves: #394 

Apply validation for MOCATOTS to be the sum of questions  1g-1l, 1n-1t, and 1w-1bb when these questions are all completed (i.e all questions described have a value of < 95)

image

### Sub question property names (1g-1l, 1n-1t, and 1w-1bb):
MOCATRAI,
MOCACUBE,
MOCACLOC,
MOCACLON,
MOCACLOH,
MOCANAMI,
MOCADIGI,
MOCALETT,
MOCASER7,
MOCAREPE,
MOCAFLUE,
MOCAABST,
MOCARECN,
MOCAORDT,
MOCAORMO,
MOCAORYR,
MOCAORDY,
MOCAORPL,
MOCAORCT

### Review checklist
- [x] Validation for MOCATOTS applies to both in person and remote/video modalities
- [x] When MOCATOTS value does not equal the sum of the values of sub questions and error will return
- [x] When 95 is supplied for a sub question and 88 is given to MOCATOTS then validation should allow for a finalize save
- [x] Remote/Telephone is not effected by this change and still saves

### DIFF
--- a/src/UDS.Net.Forms/Models/UDS4/C2.cs
+++ b/src/UDS.Net.Forms/Models/UDS4/C2.cs
@@ -786,12 +786,12 @@ namespace UDS.Net.Forms.Models.UDS4
 
         //MOCATOTS = sum of 1g-1l, 1n-1t, and 1w-1bb when all these sub questions are < 95
         [NotMapped]
-        [RequiredOnFinalized(ErrorMessage = "If questions 1g-1l, 1n-1t, and 1w-1bb are all completed, the Total Raw Score - Uncorrected needs to be the sum of these answers.")]
+        [RequiredIfInPersonVisit(nameof(MOCACOMP), "1", ErrorMessage = "If questions 1g-1l, 1n-1t, and 1w-1bb are all completed, the Total Raw Score - Uncorrected needs to be the sum of these answers.")]
         public bool? MOCATOTSSumValidation
         {
             get
             {
-                if (MOCATOTS.HasValue)
+                if (MOCATOTS.HasValue && MOCATOTS != 88)
                 {
                     int questionsSum = 0;
                     int questionsCount = 0;
--- a/src/UDS.Net.Forms/Models/UDS4/C2.cs
+++ b/src/UDS.Net.Forms/Models/UDS4/C2.cs
@@ -823,7 +823,7 @@ namespace UDS.Net.Forms.Models.UDS4
                     if (questionsCount == 19)
                     {
                         return MOCATOTS.Value == questionsSum ? true : null;
-                    }  
+                    }
                 }
 
                 return true;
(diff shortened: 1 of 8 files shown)


src/UDS.Net.Forms/Models/UDS4/C2.cs
```diff
@@ -784,6 +784,52 @@ public bool? REYDRECValidation
             }
         }
 
+        //MOCATOTS = sum of 1g-1l, 1n-1t, and 1w-1bb when all these sub questions are < 95
+        [NotMapped]
+        [RequiredOnFinalized(ErrorMessage = "If questions 1g-1l, 1n-1t, and 1w-1bb are all completed, the Total Raw Score - Uncorrected needs to be the sum of these answers.")]
```
ashleybot, line 789: @mlan225 There are some annotations designed to make clear when a property should be assessed for the visit modality. The solution in this PR certainly works, but it would be better to use this annotation because it is self-documenting. 

Something like:

```
[RequiredIfInPersonVisit(nameof(MOCACOMP), "1", ErrorMessage = "If questions 1g-1l, 1n-1t, and 1w-1bb are all completed, the Total Raw Score - Uncorrected needs to be the sum of these answers.")]
```

And then within the conditional statement, you'd check not only that `MOCATOTS.HasValue`, but also that `MOCATOTS.Value != 88`.

mlan225: Oh, thanks for pointing that annotation out I must have totally missed it. I'll get that adjusted

ashleybot: MOCACOMP determines if the section is enabled and if MOCATOTS should be evaluated.

src/UDS.Net.Forms/Models/UDS4/C2.cs
```diff
@@ -784,6 +784,52 @@ public bool? REYDRECValidation
...
+            get
+            {
+                if (MOCATOTS.HasValue && MOCATOTS != 88)
+                {
+                    int questionsSum = 0;
+                    int questionsCount = 0;
+
+                    if (MODE == FormMode.InPerson || RMMODE == RemoteModality.Video)
+                    {
+                        if (MOCATRAI.HasValue && MOCATRAI.Value < 95) { questionsCount++; questionsSum += MOCATRAI.Value; }
```
mlan225, line 801: In this conditional, there's another rule that essentially makes the check for a < 95  value not necessary for functionality. 

> If any of the sub question values are 95 - 98, then MOCATOTS must be 88

This will mean that a condition where a sub question value is 95 - 98 **should** never be reached because MOCATOTS will need to be 88 in the first place if any of these values are within that range or it will fail validation.

I figured i'd keep it in since it provides extra context on this individual validation without depending on a separate validation rule to succeed

mlan225:
### Adjustments
- [ ] Replace `Requiredonfinalize` with `RequiredIfInPersonVisit`
- [ ] Add a check for MOCATOTS != 88 before continuing with calculations

Issue #306, opened 2025-03-11
On packet submission, Errors returned, update packet status and error count


ashleybot:
![Image](https://github.com/user-attachments/assets/76465949-16bc-4dbc-aabd-fa388409ac1d)

Provide ability to click "Errors returned" and log the number of errors and update the packet status to FailedErrorChecks

![Image](https://github.com/user-attachments/assets/57f363dd-05c3-4137-b39c-c0f1eba4eb8e)

ashleybot:
When you upload a file to the NACC portal, there are two steps:
1. Three checks: form screening, file validator, identifier lookup (preliminary step to make sure the data is there and some prescreening checks)
2. Four checks: Identifier lookup, Form CSV to JSON transform, Form QC 

Whenever a packet fails any of the steps, you'll be able to see it in the portal. Status per upload. If you have a failed status and you click on it, the window opens showing which step it failed on and a link to show the errors and to resolve them.

If you have a packet that has failed and it gives you the option to "Resolve Errors" , additionally there is no way to get there unless you have a packet that has failed.

### Dynamic tabs on error view

There are different tabs at the top "File Validator " and "Form QC Checker" "Form CSV to JSON Transformer" "Form QC Checker" "Identifier Lookup" The tabs are related to which steps current have _active errors_ or _unapproved alerts_.

If there are any of those errors that exist in any of the submissions that you uploaded then the tabs will be displayed.

### Alerts

A history of all the alerts for all submissions are in this view, by default the approved alerts are hidden. By default, it shows non-approved alerts and unresolved errors. "Show approved alerts" toggle button.

You must approve alerts or resolve for the packet to be successfully submitted.

### Download/export of alerts and errors

Returns all alerts and errors for all submissions. And there is an approved columns with TRUE/FALSE.

PR #397, opened 2025-04-23
Packet Submission Visit PassedErrorChecks and FailedErrorChecks Initial Workflow

mlan225:
Resolves: #306 

Add in packet submission errors returned and success options to the packet submission UI. Early stage functionality, and will most likely undergo some further adjustments as continue to use the feature.

image

### Getting a packet to the success or errors returned step
1 -  A packet will begin in the **Pending** state until a packet has all the required forms completed
2 - Upon finalizing all required forms the packet can be finalized from visit details view (viewing all the forms in a packet). This will set the packet to the **Finalized** state
3 - Once a packet is in the submitted state, it will appear in the packet submissions details for a packet. Click view for the packet you wish to see and from here you can create a new packet submission for that packet, setting the packet to the **Submitted** state
4 - Once a packet submission is created, you can select "Success" or "Uplaod errors" depending on what was returned from the NACC upload

#### packet submission success
5a - If there were no errors returned, select "Success" to automatically set the submission to have 0 errors and change the packet status to "Passed error checks". From here, if you need to make another submission, you will need to go back to the visit and make an edit within a form in the packet, which will allow you to refinalize the packet, setting it back to the **Finalized** state. In the finalized state, you can create another submission and then mark as "Success" or "errors returned". 

#### Packet submission errors returned
5b - If there were errors, you can state the error count by selecting "Errors returned" in the packet submission view which will then display the error count input within the table

5b2 - Selecting "update error count" will set the error count for the packetsubmission and disable the buttons for further interaction. This will set the packet status to "Failed error checks" and allow you to go back and refanlize the packet

image
> This graph does not account for frozen packets 

## Feature checklist
- [x] Packet submission "success" flow sets error count to 0 and changes packet status to "PassedErrorChecks"
- [x] packet submission "errors returned" flow sets error count to user input and changes packet status to "FailedErrorChecks"
- [ ] ~~packets with "FailedErrorChecks" status can be refinalized in the packet details view~~
- [ ] ~~packets with "PassedErrorChecks" status can be refinalized in the packet details view~~
- [ ] input for Errors returned must be > 0 to be a valid input 
- [x] using the cancel button on the errors returned input will return the user to the packet submissions index view

### DIFF
--- a/src/UDS.Net.Forms/Pages/PacketSubmissions/_Edit.cshtml
+++ b/src/UDS.Net.Forms/Pages/PacketSubmissions/_Edit.cshtml
@@ -4,23 +4,40 @@
 
 
 <turbo-frame id="EditPacketSubmission">
-    <div>
-        <form method="post" asp-page="../PacketSubmissions/Edit">
-            <h1 class="text-xl">Update packet with failed status?</h1>
-            <div class="pt-3">
-                <label class="">Errors Count</label>
-                <input asp-for="ErrorCount" type="number" min="1" class="block w-full max-w-lg rounded-md border-gray-400 shadow-sm focus:border-indigo-500 focus:ring-indigo-500 sm:max-w-xs sm:text-sm placeholder:text-gray-400 disabled:bg-slate-50 disabled:text-slate-500 disabled:border-slate-200 disabled:shadow-none" />
-                <input type="hidden" asp-for="PacketId" />
-                <input type="hidden" asp-for="Id" />
-            </div>
-            <div class="pt-3 ">
-                <button type="submit" class="rounded-md bg-indigo-600 px-3 py-2 text-sm font-semibold text-white shadow-sm hover:bg-indigo-500 focus-visible:outline focus-visible:outline-2 focus-visible:outline-offset-2 focus-visible:outline-indigo-600">
-                    Yes, Update
-                </button>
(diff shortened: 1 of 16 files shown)


mlan225:
Something that we already mentioned a little bit before was that this button here does not refresh after changing packet status. This is because turbo used around the packet submissions logic does not refresh the parent page. Last time we spoke about it, it didn't seem to be a priority as it does not interrupt Josh's use case, in the future there may be some turbo related things we can do to force a reload or even just redirect to the view if we wanted to force it that way 🤷 

At the moment, it will gray out appropriately on refresh, but may cause some issues if clicked out of the main pipeline.

image

src/UDS.Net.Forms/Pages/PacketSubmissions/_Edit.cshtml
```diff
@@ -0,0 +1,43 @@
...
+                <tr>
+                    <td colspan="2">
+                        <p class="text-center">Update submission error count and set failed status?</p>
+                    </td>
+                    <td>
+                        <label class="">Errors Count</label>
+                        <input asp-for="ErrorCount" type="number" min="1" class="block w-full max-w-lg rounded-md border-gray-400 shadow-sm focus:border-indigo-500 focus:ring-indigo-500 sm:max-w-xs sm:text-sm placeholder:text-gray-400 disabled:bg-slate-50 disabled:text-slate-500 disabled:border-slate-200 disabled:shadow-none" />
+                        <input type="hidden" asp-for="PacketId" />
+                        <input type="hidden" asp-for="Id" />
+                        <input type="hidden" name="packetStatus" value="3" />
```
ashleybot, line 27: @mlan225 Something that you can use to identify where things should be improved, or a "code smell", is when things are hard-coded. Having this `packetStatus` hard-coded as 3 will cause long-term maintenance difficulties. But I don't want you to change it now. Get it merged, version bumped, and into our local codebase and we will pair in our 1:1 and I'll show you alternatives.

PR #401, opened 2025-04-30
D1b new fdgpet enhanced validation

### LINKED ISSUE #345: D1b New FDGPET enhanced validation
Add a new NotMapped property for enhanced validation for the FDGPET section

### Specification

If FDGAD, FDGFTLD, FDGLBD, FDGOTH are NOT NULL then at least one must not be 8.

- [ ] Add the NotMapped property
- [ ] Display the error message in a proximal location to the FDG table
- [ ] Extensively test all failure and success states including correct failures

### One example of a success state

Image

### Reference

- [NACC specifications line 71](https://github.com/naccdata/uniform-data-set/blob/main/forms/uds/d1b/form_d1b_ivp_error_checks_mc.csv)

CallowaySimon:
Resolves #345 

## D1b New FDGPET enhanced validation
Adds a new NotMapped property for enhanced validation for the FDGPET section

image

### Review Checklist

- [ ] Validation error if user tries to finalize form while FDGPETDX == 1 or 2 and FDGAD, FDGFTLD, FDGLBD, FDGOTH == 8
- [ ] Validation error displays under FDGOTH

### DIFF
--- a/src/UDS.Net.Forms/Models/UDS4/D1b.cs
+++ b/src/UDS.Net.Forms/Models/UDS4/D1b.cs
@@ -135,6 +135,24 @@ namespace UDS.Net.Forms.Models.UDS4
         [Display(Name = "FDG PET - Was FDG PET data or information used to support an etiological diagnosis?")]
         public int? FDGPETDX { get; set; }
 
+        [NotMapped]
+        [Display(Name = "FDG Etiology validation")]
+        public bool? FDGEtiologyValidation
+        {
+            get
+            {
+                if (FDGAD != null && FDGFTLD != null && FDGLBD != null && FDGOTH != null)
+                {
+                    if (FDGAD != 8 || FDGFTLD != 8 || FDGLBD != 8 || FDGOTH != 8)
+                    {
+                        return true;
+                    }
+                    return false;
+                }
+                return null;
+            }
+        }
+
         [RequiredIfRange(nameof(FDGPETDX), 1, 2, ErrorMessage = "Please specify.")]
         [Display(Name = "Consistent with AD")]
         public int? FDGAD { get; set; }
@@ -636,6 +654,14 @@ namespace UDS.Net.Forms.Models.UDS4
                     "At least one CSF-based biomarker must not be 8.",
                     new[] { nameof(CSFOTH) }
                 );
+            }
+
+            if(FDGEtiologyValidation == false)
+            {
+                yield return new ValidationResult(
+                    "At least one FDG etiology must must not be 8",
+                    new[] { nameof(FDGOTH)}
+                    );
             }
 
             yield break;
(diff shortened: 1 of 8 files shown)


src/UDS.Net.Forms/Models/UDS4/D1b.cs
```diff
@@ -638,8 +656,15 @@ public override IEnumerable<ValidationResult> Validate(ValidationContext validat
                 );
             }
 
+            if (FDGEtiologyValidation == false)
```
ashleybot, line 659: @CallowaySimon Typically, we only want to use this section if the validation is too complex to have only as a NotMapped property, but not both.

I see what you might have been doing to display the error message, but in this case we want to modify the markup and add another tag for `asp-validation-for="FDGEtiologyValidation"`

src/UDS.Net.Forms/Models/UDS4/D1b.cs
```diff
@@ -135,6 +135,24 @@ public bool? SPECTValidation
         [Display(Name = "FDG PET - Was FDG PET data or information used to support an etiological diagnosis?")]
         public int? FDGPETDX { get; set; }
 
+        [NotMapped]
+        [Display(Name = "FDG Etiology validation")]
+        public bool? FDGEtiologyValidation
```
ashleybot, line 140: @CallowaySimon Add annotations for `RequiredIf` watching the FDGPETDX. This doubles as self-documenting code!

src/UDS.Net.Forms/Models/UDS4/D1b.cs
```diff
@@ -135,6 +135,24 @@ public bool? SPECTValidation
...
+        {
+            get
+            {
+                if (FDGAD != null && FDGFTLD != null && FDGLBD != null && FDGOTH != null)
+                {
+                    if (FDGAD != 8 || FDGFTLD != 8 || FDGLBD != 8 || FDGOTH != 8)
+                    {
+                        return true;
```
ashleybot, line 148: @CallowaySimon And the conditional will need to be adjusted to return null in the fail state and true in the success (or does not need to be validated) state. See `BloodBiomarkersValidation` for an example.

ashleybot (changes requested):
@CallowaySimon There are some standards to follow with validation that you don't yet know about because you are new to this codebase.

src/UDS.Net.Forms/Models/UDS4/D1b.cs
CallowaySimon: Removed section from Validate and added span for error message above radio buttons 
image

src/UDS.Net.Forms/Models/UDS4/D1b.cs
CallowaySimon: Added `RequiredIf` annotations
image

src/UDS.Net.Forms/Models/UDS4/D1b.cs
CallowaySimon: Set to return True in Success state a null in Fail state
image

PR #434, opened 2025-05-30
UI behavior changed

### LINKED ISSUE #433: A5D2 When SEIZURES = 2 then SEIZNUM should be disabled
Rules for question 4c `SEIZURES` are slightly different than what we have setup in UI behaviors.

Image

Update UI Behaviors and RequiredIf logic so that these rules are enforced:

- [ ] When SEIZURES = 2, SEIZNUM should be disabled and not required, SEIZAGE should be enabled and required
- [ ] When SEIZURES = ~~1~~ 0 or 9, SEIZNUM should be disabled and not required, SEIZAGE should be disabled and not required.
- [ ] When SEIZURES = 1, SEIZNUM should be enabled and required, SEIZAGE should be enabled and required

EDITED to be complete specifications.

## Reference
https://files.alz.washington.edu/documentation/uds4-ivp-a5-d2.pdf

ahmedhaly7777:
Fixes #433 

**Side fix:** An error message appears for SEIZURES even when it’s disabled, SEIZURES equals 2, and SEIZAGE is not empty. There should be no error message when SEIZURES is disabled.

### DIFF
--- a/src/UDS.Net.Forms/Models/UDS4/A5D2.cs
+++ b/src/UDS.Net.Forms/Models/UDS4/A5D2.cs
@@ -168,7 +168,6 @@ namespace UDS.Net.Forms.Models.UDS4
         public int? SEIZURES { get; set; }
         [Display(Name = "How many seizures has the participant had in the past 12 months?")]
         [RegularExpression("^([0-2]|9)$", ErrorMessage = "Valid range is 0-2 or 9")]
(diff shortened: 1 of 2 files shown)


ashleybot (changes requested):
@ahmedhaly7777 SEIZNUM should still be required and enabled when SEIZURES = 1

Issue #436, opened 2025-06-03
B6 calculation bug fix


mlan225:
### Description
The calculations for the B6 form counted radio values of "9" as answered questions. This caused the prorated calculations to be off and caused an error in a submission of a packet when a GDS score was not what NACC expected. 

### Steps to reproduce
1. Go to B6 form
2. Fill in the following values for questions 1 - 15 (view screenshot for visual representation)

```
1. 1 
2. 1 
3. 1 
4. 1 
5. 0 
6. 1 
7. 0 
8. 1 
9. 0 
10. 0 
11. 9 
12. 0 
13. 1 
14. 0 
15. 0
```

### Expected behavior
The answer should be rounded to 8, but is currently rounding down to 7, since the non-rounded result is 7.4 instead of 7.5

### Screenshots
![Image](https://github.com/user-attachments/assets/81352ba6-fbe5-446b-98bd-a9066a45c999)

PR #437, opened 2025-06-03
Address bug causing incorrect scores in some scenarios

mlan225:
Resolves: #436 

### Address bug with B6 calculations

The issue was that checked radios with value of "9" were being counted as "answered" radios and being used towards the prorate calculations. Now, the count for answered and unanswered radios are tracked separately so unanswered radios can be correctly calculated in the result.

Filling out the form with the values now displays the correct result

```
1. 1 
2. 1 
3. 1 
4. 1 
5. 0 
6. 1 
7. 0 
8. 1 
9. 0 
10. 0 
11. 9 
12. 0 
13. 1 
14. 0 
15. 0
```

image

Checklist:
- [ ] Filling out the values in the above screenshot results in a value of 8
- [ ] filling out the results with the calculations results in the appropriate decimal value to be rounded using the formula:
>  **RESULT IS ROUNDED** 
> Total score of completed items + ((total score of completed items / # of completed items) * (# of unanswered items))

in the above example the prorated score will be:
> 7 + ((7 / 14) * (1)) = 7.5, which is then rounded to 8


ahmedhaly7777 (approved):
LGTM!

ahmedhaly7777:
@ashleybot Ready for your review!

ashleybot (changes requested):
@mlan225 Does this take into account when 3 or more questions are 9 (Did Not Answer)? I selected more than 3 questions as 9 and GDS was not set to 88.

Reference => https://github.com/UK-SBCoA/uniform-data-set-dotnet-web/pull/263

mlan225:
@ashleybot at the moment the validation will want the user to manually select the checkbox when more than 3 questions are 9 or provide an error message upon finalize

image

Do we want this to just check the box and set the value automatically or leave with this current validation functionality?

mlan225:
### Marvin updates
- [ ] Added code to the JS that will auto check NOGDS and fill 88 for GDS score when more than 3 radios are given a value of 9

Originally the validation would wait for a finalized submission before displaying an error, but now the radios will handle the NOGDS functionality automatically while going through the radio buttons. 

The checkbox is still clickable, and will also autofill GDS score to 88, but if a value of 88 is submitted without more than 3 radios having a value of 9, an error validation message will occur and stop submission.

mlan225:
Updated my conversation wtih Justin on the parent card. Looks like his update didn't warrant any code changes to the recent updates. 

I'll move this back into review and we can get together and regroup on all the information when we get around to having this reviewed @ashleybot

ashleybot (changes requested):
@mlan225 Js is looking much cleaner and easier to understand the rules. 

Is there a way to keep the more self-documenting RequiredIf annotation? The radio buttons should not need to be required if NOGDS is true. 

> Check this box and enter "88" if and only if the participant: 1.) does not attempt the GDS, or 2.) answers fewer than 12 questions.

Issue #439, opened 2025-06-04
C2 enhanced validation (UDSVERTE, UDSVERTN, MINTSCNC)


mlan225:
Packet submission revealed the following errors that are able to be saved on forms

- [ ]  if q11b. udsverfn (f-words repeated) is between 0 and 15 and q11e. udsverlr (l-words repeated) is between 0 and 15, then q11h. udsverte must be the total of q11b. udsverfn and q11e. Udsverlr
- [ ]  if q11a. udsverfc not 95-98 and q11d. udsverlc not 95-98, q11g. udsvertn must be the total of udsverfc and udsverlc

![Image](https://github.com/user-attachments/assets/d54eb35f-18f3-440e-bbd1-b9eece430b34)

- [ ]  if q16c. mintscng (mint number semantic cues given) is > 0 then q16d. mintscnc (mint correct with semantic cue) must be less than or equal to q16c. mintscng (mint number semantic cues given)

![Image](https://github.com/user-attachments/assets/bbaa7828-2671-4348-871b-6cd51dd5cdfe)

PR #442, opened 2025-06-04
C2 enhanced validation

mlan225:
Resolves: #439 

### Add validation
The 3 validation rules were added and will reveal an error message when the conditions are met

#### These questions refer to both in-person/video & telephone (the numbering may be different)
- [x]  if q11b. udsverfn (f-words repeated) is between 0 and 15 and q11e. udsverlr (l-words repeated) is between 0 and 15, then q11h. udsverte must be the total of q11b. udsverfn and q11e. Udsverlr
- [x]  if q11a. udsverfc not 95-98 and q11d. udsverlc not 95-98, q11g. udsvertn must be the total of udsverfc and udsverlc

#### This question is specific to the in-person/video mode
- [x]  if q16c. mintscng (mint number semantic cues given) is > 0 then q16d. mintscnc (mint correct with semantic cue) must be less than or equal to q16c. mintscng (mint number semantic cues given)

image

image


src/UDS.Net.Forms/Models/UDS4/C2.cs
```diff
@@ -830,6 +830,63 @@ public bool? MOCATOTSSumValidation
...
+        }
+
+        [NotMapped]
+        [RequiredIfRange(nameof(MINTTOTS), 0, 32, ErrorMessage = "If MINTSCNG (mint number semantic cues given) is > 0 then MINTSCNC (mint correct with semantic cue) must be less than or equal to MINTSCNG (mint number semantic cues given")]
+        public bool? MINTSCNCValidation
+        {
+            get
+            {
+
+                if ((MINTSCNG.HasValue && MINTSCNG.Value > 0) && (MINTSCNC.HasValue && MINTSCNC.Value != 88) && (RMMODE != RemoteModality.Telephone))
```
ashleybot, line 876: @mlan225 @Oddvocado These conditionals could be formatted differently for improved readability and code performance. If the RMMODE is telephone then nothing else should be assessed, so it makes sense for it to be a top level conditional (improves performance).
```
if (RMMODE != RemoteModality.Telephone)
{
}
return true;
```

Next, the statement is not assessable if all properties do not have a value. So, that should go next. As a reader, this clearly sets up the situation and I know exactly which variables are being evaluated.
```
if (RMMODE != RemoteModality.Telephone)
{
    if (MINTSCNG.HasValue && MINTSCNC.HasValue)
    {
    }
}
return true;
```

Then finally, the rule being assessed in as simple as a statement as possible. If MINTSCNC != 88, then MINTSCNC _must_ be less than MINTSCNG. Having this line in one conditional makes it incredibly simple to understand the rule. I've also used the opposite of the rule here so that the failing scenario is clear and it reduces the number of return lines.
```
if (RMMODE != RemoteModality.Telephone)
{
    if (MINTSCNG.HasValue && MINTSCNC.HasValue)
    {
        if (MINTSCNC.Value != 88 && MINTSCNC.Value > MINTSCNG.Value)
        {
            return null; // MINTSCNC cannot be larger than MINTSCNG
        }
    }
}
return true;
```

ashleybot (changes requested):
@mlan225 Let's make this code more easily maintainable by 

- [ ] Adjusting conditional statements for readability and performance
- [x] And moving these properties up higher in proximity to other properties they are related to, so for example, MINT property will move up higher to Line 360-ish

@mlan225 @Oddvocado Generally, if/else conditional statements are more readable if you create a logical hierarchy and grouping of "sameness." So, this code would be more readable if all the checks for if relevant properties have a value are in the same conditional statement. Likewise, anything that would invalidate the whole statement and reduce assessment by the compiler should go first. (see comment).

While reviewing and testing I noticed another MINT validation statement that really should be another unmapped validation property. It would be good to adjust this along with this feature.

- [x] Move MINTTOTS, MINTTOTW, MINTSCNC validation into its own validation property, on lines 1011-1023
- [x] MINTTOTS also has required if validation annotations, but it is not clear if this works as intended or if the required validation needs to remain in the Validate() method, line 1006. The question is if required validation is completely handled by annotations, and if they are correct, or if it is unnecessarily duplicated.

mlan225:
### Updates
- [ ] Adjusting conditional statements for readability and performance
- [ ] And moving these properties up higher in proximity to other properties they are related to, so for example, MINT property will move up higher to Line 360-ish
- [ ] Move MINTTOTS, MINTTOTW, MINTSCNC validation into its own validation property, on lines 1011-1023
- [ ] removed attribute validations [RequiredIf(nameof(MODE)] and [RequiredIf(nameof(RMMODE)]. 
  - After testing these validation attributes, they did not appear to actually validate correctly, and the logic was already indeed covered in another section of the code

mlan225:
Updated the UDSVERTN and UDSVERTE validation to account for 95 - 98 "other" values.

### Example of fix
UDSVERTN validation relies on UDSVERFC and UDSVERLC to both be 0 - 40 and not 95 - 98. Originally, if one of these values were 95 - 98, then the validation would fail, which is not intended. Now, UDSVERTN validation will not attempt to validate if the pre-reqresets are not met (UDSVERFC and UDSVERLC both values of 0 - 40) 

The same logical fix was applied to UDSVERTE validation

image

mlan225:
Trail B validation added to PR, we can talk about this in stand up

ashleybot (approved):
@mlan225 Good to merge and bump patch. Not necessary to update COA.UDSv4, we'll do that after @Oddvocado's PR is done.

PR #445, opened 2025-06-10
Validation changed

### LINKED ISSUE #292: Update D1a Question 5b IMPNOMCI js
Update D1a Question 5b to use UI Behaviors instead of custom jquery

`IMPNOMCI` should disable Questions 6a-6g.

# Custom jquery from D1a.js should be removed

```
    function updateIMPNOMCITargets() {
        var selectedValue = $("input[name='D1a.IMPNOMCI']:checked").val();
        var shouldDisable = (selectedValue == '0' || selectedValue == '1');
        var IMPNOMCITargets = [
            "D1a.CDOMMEM", "D1a.CDOMLANG", "D1a.CDOMATTN", "D1a.CDOMEXEC", "D1a.CDOMVISU", "D1a.CDOMBEH", "D1a.CDOMAPRAX"
        ];

        IMPNOMCITargets.forEach(function (inputName) {
            $('input[name="' + inputName + '"]').prop('disabled', shouldDisable);
        });
    }

    updateIMPNOMCITargets();

    $("input[name='D1a.IMPNOMCI']").change(updateIMPNOMCITargets);
```

ahmedhaly7777:
Fixes #292

### DIFF
--- a/src/UDS.Net.Forms/Pages/UDS4/D1a.cshtml.cs
+++ b/src/UDS.Net.Forms/Pages/UDS4/D1a.cshtml.cs
@@ -302,6 +302,34 @@ namespace UDS.Net.Forms.Pages.UDS4
             } }
         };
 
+        public Dictionary<string, UIBehavior> IMPNOMCIUIBehavior = new Dictionary<string, UIBehavior>
+        {
+            { "0", new UIBehavior {
+                PropertyAttributes = new List<UIPropertyAttributes>
+                {
+                    new UIDisableAttribute("D1a.CDOMMEM"),
+                    new UIDisableAttribute("D1a.CDOMLANG"),
+                    new UIDisableAttribute("D1a.CDOMATTN"),
+                    new UIDisableAttribute("D1a.CDOMEXEC"),
+                    new UIDisableAttribute("D1a.CDOMVISU"),
+                    new UIDisableAttribute("D1a.CDOMBEH"),
+                    new UIDisableAttribute("D1a.CDOMAPRAX")
+                }
+            } },
+            { "1", new UIBehavior {
+                PropertyAttributes = new List<UIPropertyAttributes>
+                {
+                    new UIDisableAttribute("D1a.CDOMMEM"),
+                    new UIDisableAttribute("D1a.CDOMLANG"),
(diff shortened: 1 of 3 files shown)


ahmedhaly7777:
Change UI behavior so that Q3 Demented & Q4b MCI use UI behaviors to enable and disable Q6a-Q6g.

ashleybot (changes requested):
@ahmedhaly7777 We're getting there. Nice work!

These are true:
- [x] New form loads in correct state with Q6 checkboxes disabled
- [x] Q6 correctly enables when DEMENTED == 1
- [ ] Q6 correctly disables when DEMENTED == 0
- [ ] Q6 correctly disables when MCI == 1
- [ ] Q6 correctly disables when MCI == 0

However, if I change my mind and back up after selected a value for MCI and change DEMENTED again then I can reach a state where the checkboxes are not correctly enabled again.

Steps to repeat:
1. NORMCOG == 0
2. DEMENTED == 0
3. MCI == 1
4. Checkboxes are correctly enabled
7. Back up to Q3 DEMENTED = 1
8. Checkboxes are correctly enabled
9. The state for Q4 is persisted and MCI still == 1, but js is not ran for enabled children so the checkboxes are _incorrectly disabled_

So, we either need to iterate the generations of enabled children with values, or clear values on disable. Both are wide-breadth changes. Can you think through pros/cons of each approach and the effects on all forms and we'll discuss in our 1:1?

src/UDS.Net.Forms/Pages/UDS4/D1a.cshtml.cs
```diff
@@ -302,6 +330,34 @@ public D1aModel(IVisitService visitService, IParticipationService participationS
             } }
         };
 
+        public Dictionary<string, UIBehavior> IMPNOMCIUIBehavior = new Dictionary<string, UIBehavior>
```
ashleybot, line 333: @ahmedhaly7777 I don't think I encountered a scenario where this is necessary. IMPNOMCI is only enabled when MCI == 0 and that UI Behavior handles disabling on Q6.

ahmedhaly7777:
@ashleybot Ready for review!

PR #454, opened 2025-06-17
C2 In-Person Validation

### LINKED ISSUE #447: Fix C2 In-Person Fields Validation
Error messages for the following fields are not displayed when left empty upon finalizing the form:

~~8b `TRAILB`~~
~~13a `REYDREC`~~
8b `TRAILB` -- See comment below
15d `CERADJ7YES`
15e `CERADJ7NO`

Oddvocado:
Resolves #447

### DIFF
--- a/src/UDS.Net.Forms/Models/UDS4/C2.cs
+++ b/src/UDS.Net.Forms/Models/UDS4/C2.cs
@@ -569,10 +569,12 @@ namespace UDS.Net.Forms.Models.UDS4
 
         [Display(Name = "J7 Word List Recognition: Total YES correct", Description = "(0-10,95-98)")]
         [RegularExpression("^(\\d|10|9[5-8])$", ErrorMessage = "Allowed values are 0-10 or 95-98.")]
+        [RequiredIfRange(nameof(CERADJ6REC),0,10, ErrorMessage = "Response required")]
         public int? CERADJ7YES { get; set; }
 
         [Display(Name = "J7 Word List Recognition: Total NO correct", Description = "(0-10,95-98)")]
         [RegularExpression("^(\\d|10|9[5-8])$", ErrorMessage = "Allowed values are 0-10 or 95-98.")]
+        [RequiredIfRange(nameof(CERADJ7YES), 0, 10, ErrorMessage = "Response required")]
         public int? CERADJ7NO { get; set; }
 
         [Display(Name = "Part A: Total number of seconds to complete", Description = "(0-100, 888, 995-998)")]


src/UDS.Net.Forms/Models/UDS4/C2.cs
```diff
@@ -569,10 +569,12 @@ public class C2 : FormModel
 
         [Display(Name = "J7 Word List Recognition: Total YES correct", Description = "(0-10,95-98)")]
         [RegularExpression("^(\\d|10|9[5-8])$", ErrorMessage = "Allowed values are 0-10 or 95-98.")]
+        [RequiredIfRange(nameof(CERADJ6REC), 0, 10, ErrorMessage = "Response required")]
```
ashleybot, line 572: @Oddvocado This is an opportunity to add enhanced error checking for the CERAD section, seems that Total YES correct + Total No correct should equal 10, if not 95-98. Could you look into the documentation?

ashleybot (changes requested):
@Oddvocado Merged main into your feature branch and wondering about documentation.

src/UDS.Net.Forms/Models/UDS4/C2.cs
ashleybot: @Oddvocado More documentation here for C2 https://naccdata.org/data-collection/forms-documentation/uds-4

Oddvocado:
Validation is working on my end as well! I tested all the inputs that are now using the [RequiredIfInPersonVisit] annotation and the validation is being displayed and preventing form submission as expected. Both in-person and remote/video scenarios were tested. 

Ready for another review @ashleybot !

ashleybot (changes requested):
@Oddvocado The custom annotation needs to have the AddValidation() method updated so that the data attributes are rendered properly.

**Incorrectly rendered data attributes:**
Screenshot 2025-07-21 at 8 17 33 AM

> @ashleybot While investigating the bug for the property TRAILB, I discovered that the custom data annotation was incorrectly labeled as: `[RequiredIf(nameof(MODE), "In-Person")]`
> 
> When it should have been
> 
> `[RequiredIf(nameof(MODE), "InPerson")]`
> 

Modifying the annotation is an ok solution, that said, I wanted to make sure that you know this would not work correctly. The values for MODE and RMMODE drop-down are integer values. So it would need to be `[RequiredIf(nameof(MODE), "1")]`

src/UDS.Net.Forms/DataAnnotations/RequiredIfInPersonVisitAttribute.cs
```diff
@@ -30,49 +37,64 @@ public RequiredIfInPersonVisitAttribute(string watchedField, string value) : bas
...
         }
 
         public void AddValidation(ClientModelValidationContext context)
         {
             MergeAttribute(context.Attributes, "data-val", "true");
             MergeAttribute(context.Attributes, "data-val-requiredif", this.ErrorMessage);
-            string watched = _watchedField;
-            var containerName = context.ModelMetadata.ContainerType.Name;
-            if (!String.IsNullOrEmpty(containerName))
+            if (!_inPersonCheckOnly)
```
ashleybot, line 87: @Oddvocado You've removed the watched data attributes, but does the requiredif client code work properly to support client side validation?

If we don't render any data attributes when the field is required, does client side validation happen at all? You could utilize the built-in required validator.

So instead of rendering data-val-required-if, you'd use data-val-required.

When there is no watched field add this data attribute:
```
MergeAttribute(context.Attributes, "data-val-required", this.ErrorMessage);
```

Something like this would be rendered for those inputs:
`<input ... data-val="true" data-val-required="This is required" />`

ashleybot (changes requested):
@Oddvocado There's some opportunity for code readability improvements. And more needs to be done with rendering the correct data attributes.

Oddvocado:
@ashleybot ready for another review! I’ve updated the conditional logic so that the correct data-* attributes are applied based on the value of _inPersonCheckOnly.

Looking at the developer tools in chrome, when  _inPersonCheckOnly set to true, only the data-val-required attribute is added, both data-val-requiredif attributes are omitted.

image

image

Additionally, for C2 properties whose validation depends on the value of another property ( _inPersonCheckOnly is set to false), the appropriate data watch field attributes are now correctly applied.

image
image



Per our conversation earlier today, there still appear to be some issues with unobtrusive validation handling regular expressions. Additionally, the C2 form is not recognizing jQuery syntax under certain conditions. I’m currently working on reproducing the steps to reliably trigger the jQuery issue so I can document it in a new card.

src/UDS.Net.Forms/DataAnnotations/RequiredIfInPersonVisitAttribute.cs
```diff
@@ -10,17 +10,24 @@
...
 namespace UDS.Net.Forms.DataAnnotations
 {
+    // RequiredIfInPersonVisit is primarily used in the C2 form. "InPerson" refers to both in-person visits and remote/video visits.
+
     [AttributeUsage(AttributeTargets.Property, AllowMultiple = true)]
     public class RequiredIfInPersonVisitAttribute : ValidationAttribute, IClientModelValidator
     {
         private string _watchedField = "";
         private string _watchedFieldValue = "";
+        private readonly bool _inPersonCheckOnly = false;
```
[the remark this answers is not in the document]
Oddvocado: @ashleybot I've gone ahead and removed '_inPersonCheckOnly ' and replaced it with a string.IsNullOrEmpty comparison.

Issue #451, opened 2025-06-17
In export, if MODE = 0 set fields to null


ashleybot:
Columns to set to NULL rather than using the value:
FRMDATE
INTIALS
LANG
ADMIN

PR #455, opened 2025-06-18
Handle record properties for non complete

mlan225:
Fixes: #451 

### Handle null values expected for non complete forms in export
- When a non complete reason code is detected, NACC expects the `mode` and `not` variables to be THE ONLY fields with values, everything else for the form should be null in the export. This requires us to null out values in the record and formfields for the export (but retain the data in our systems)

### Steps for testing
1. Finalize a packet
2. Create a new submission for the finalized packet
3. Select the download link to run the export 

- [ ] When providing a non complete form mode the export will have null values for every column for the form except in `mode` and `not` (i.e a2amode & a2anot)
- [ ] When a form is given a non complete mode, the export will null out previously answered questions in the export. Existing answers in the form will not display in the export with a non complete mode

#### Example of export with non complete mode (B7 form)
image


mlan225:
### Notes to self
Spoke with Ash today

- [ ] Update the FormsContract to match list of required forms from NACC, 
  - can use these requirements to check if we need to null out the export properties
- [ ] Do some more looking into seeing if I can extend the WriteRecord method to handle nulling out fields
- [ ] We will be nulling out the form fields as well
- [ ] look into having the export code moved to the record (this may be a different approach than extending writeRecord)


- a1 and a1a are the only forms that have the admin prop

mlan225:
Updated the code to add back in the Enum `IgnoredFormFieldProps`. Earlier, we tested a line of code that looked for classes that implement the interface `IEnumerable`, however, the string class implements IEnumerbale, so it would require additional reflection  include checking for string types

Issue #464, opened 2025-07-02
Update packet finalization logic


ashleybot:
Currently we only require forms to be finalized if they are required. This is not completely correct logic.

https://github.com/UK-SBCoA/uniform-data-set-dotnet-web/blob/163520663a08bef0c1615b8ee660463facff0aa1/src/UDS.Net.Services/DomainModels/Visit.cs#L81

If a form is optional (not required for visit kind) then we still should require it to be finalized, but it can be finalized with `MODE=NotCompleted` and a `NOT reason code`.

So, update the logic to require every form in the form contract to be finalized. If the form is required for visit kind then add an additional check to **not** allow MODE = NotCompleted.

PR #465, opened 2025-07-07
Update packet finalization logic

mlan225:
Resolves: #464 

Draft PR

### DIFF
--- a/src/UDS.Net.Services/DomainModels/Visit.cs
+++ b/src/UDS.Net.Services/DomainModels/Visit.cs
@@ -78,6 +78,7 @@ namespace UDS.Net.Services.DomainModels
                                 // if the form contract exists
                                 var form = this.Forms.Where(f => formContract.Abbreviation == f.Kind).FirstOrDefault();
 
+                                //Start with condition here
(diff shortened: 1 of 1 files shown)


ashleybot (changes requested):
@mlan225 That's a good category of use cases to test, but I think it needs to be renamed and then more added.

- [ ] Rename `IsFinalizable_ReturnsFalse_WhenPacketStatusNotPending` to `IsFinalizable_ReturnsFalse_WhenPacketStatusInSubmitted` and add the suite of finalized forms

Add test cases

- [ ] IsFinalizable packet status FailedErrorChecks and all forms finalized = true
- [ ] IsFinalizable packet status PassedErrorChecks and all forms finalized = true
- [ ] IsFinalizable packet status Frozen and all forms finalized = false

mlan225:
### TryUpdate based on packet status
The finalize logic has two checks: 
- IsFinalizable must be true 
- TryUpdateStatus must be true

Added an additional check to unit tests covering packet status to include TryUpdateStatus


### Update to include all forms in unit tests
- Unit tests that were checking for failure would fail automatically if all forms were not included. This would result in a false positive for the test pass

Issue #460, opened 2025-06-25
A1 Race Validation


mlan225:
### Enhanced race and ethnicity validation

![Image](https://github.com/user-attachments/assets/2aa9fb5a-4407-400c-b358-4a780cba0966)

Question 3 does not currently require you to answer a parent item if you've answered a subcategory. 

The current validation for checking ethnicity boxes is as follows: 
  - ![Image](https://github.com/user-attachments/assets/e3e40045-47ad-40e8-8196-b4b418549ff8)
    - This validation only checks that one of the checkboxes are selected and NOT if any of the parent items are selected

### NACC error that josh encountered
> At least one of the following variables must be equal to 1: Q3a. RACEAIAN, Q3b. RACEASIAN, Q3c. RACEBLACK, Q3d. ETHISPANIC, Q3e. RACEMENA, Q3f. RACENHPI, Q3g. RACEWHITE, Q3h. RACEUNKN	

### Steps to reproduce
1. Go to the A1 form
2. select at least one sub checkbox in the ethnicity and race section BUT NOT any of the parent items (in section 3)
3. Complete the rest of the form and finalize
4. The form will save successfully

### Task
- [ ] Form will need to require a parent category upon completion, even if a child is selected

ashleybot:
This validation rule is that at least one value needs to be selected from these variables:

- RACEAIAN
- RACEASIAN
- RACEBLACK
- ETHISPANIC
- RACEMENA
- RACENHPI
- RACEWHITE
- RACEUNKN

PR #472, opened 2025-07-08
A1 race validation Solution

mlan225:
Resolves: #460 

## Race validation adjustment
Race checkboxes are organized as Race and ethnicities. These groupings are consistent with "Unknown race" at the bottom

image

### Review checklist
- [ ] When unknown race is selected, a race checkbox cannot be selected
- [ ] an ethnicity can be selected without a race option selected
- [ ] an ethnicity can be selected AND unknown race can be selected
- [ ] Validation messages have been adjusted to focus on selecting race rather than race/ethnicity 
- [ ] At least one race OR unknown must be selected

### DIFF
--- a/src/UDS.Net.Forms/Models/UDS4/A1.cs
+++ b/src/UDS.Net.Forms/Models/UDS4/A1.cs
@@ -217,7 +217,7 @@ namespace UDS.Net.Forms.Models.UDS4
         [Display(Name = "Don't know")]
         public bool RACEUNKN { get; set; }
 
-        [RequiredOnFinalized(ErrorMessage = "Please indicate ethnicity/race or don't know.")]
+        [RequiredOnFinalized(ErrorMessage = "Please indicate race or don't know.")]
         [NotMapped]
         public bool? EthnicityRaceIndicated
         {
@@ -232,7 +232,7 @@ namespace UDS.Net.Forms.Models.UDS4
             }
         }
 
-        [RequiredOnFinalized(ErrorMessage = "Don't know cannot be selected along with a known ethnicity/race.")]
+        [RequiredOnFinalized(ErrorMessage = "Don't know cannot be selected along with a known race.")]
         [NotMapped]
         public bool? EthnicityRaceUnknownOnly
         {
--- a/src/UDS.Net.Forms/Models/UDS4/A1.cs
+++ b/src/UDS.Net.Forms/Models/UDS4/A1.cs
@@ -238,7 +238,7 @@ namespace UDS.Net.Forms.Models.UDS4
         {
             get
             {
-                if (RACEAIAN || RACEASIAN || RACEBLACK || ETHISPANIC || RACEMENA || RACENHPI || RACEWHITE )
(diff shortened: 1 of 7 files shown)


mlan225:
Will need to further investigate the form rules to see the correlation between the parent items and child items. This grouping of the race options may be on our end. If it is just on our end and not a NACC requirement, speak with Josh about if this should be enforced (i.e selecting a "parent" category and then being able to select a separate race "child item".

mlan225:
### From looking at the guideline rules
- a child item is not required, but at least one parent race checkbox (including unknown) is required
- A child item does not need to be selected within the checked parent item group for validation to pass, as long as at least one parent item is checked it will pass
- For unknown to be checked, a parent race item must not be checked. It does not seem like the validation rules state a child item cannot be selected, so in theory a child item + unknown can be checked and pass validation (should we allow this?)

src/UDS.Net.Forms/Models/UDS4/A1.cs
```diff
@@ -221,50 +217,33 @@ public class A1 : FormModel
...
             get
             {
-                if (RACEWHITE || ETHGERMAN || ETHIRISH || ETHENGLISH || ETHITALIAN || ETHPOLISH || ETHSCOTT || ETHWHIOTH
-                    || ETHISPANIC || ETHMEXICAN || ETHPUERTO || ETHCUBAN || ETHSALVA || ETHDOMIN || ETHGUATEM || ETHHISOTH
-                    || RACEBLACK || ETHAFAMER || ETHJAMAICA || ETHHAITIAN || ETHNIGERIA || ETHETHIOP || ETHSOMALI || ETHBLKOTH
-                    || RACEASIAN || ETHCHINESE || ETHFILIP || ETHINDIA || ETHVIETNAM || ETHKOREAN || ETHJAPAN || ETHASNOTH
-                    || RACEAIAN
-                    || RACEMENA || ETHLEBANON || ETHIRAN || ETHEGYPT || ETHSYRIA || ETHIRAQI || ETHISRAEL || ETHMENAOTH
-                    || RACENHPI || ETHHAWAII || ETHSAMOAN || ETHCHAMOR || ETHTONGAN || ETHFIJIAN || ETHMARSHAL || ETHNHPIOTH)
+                if (RACEAIAN || RACEASIAN || RACEBLACK || ETHISPANIC || RACEMENA || RACENHPI || RACEWHITE || RACEUNKN)
```
ashleybot, line 241: @mlan225 This code reads as if RACEUNKN = true validation will fail every time.
```
if (... RACEUNKN)
{
    if (RACEUNKN) return null;
}
```

mlan225: OH WOW, a last minute change on my part messed that up.

I'll get that fixed up and provide the list of tests I've ran to check

ashleybot (approved):
@mlan225 Looks good! Bump patch.

Successful tests:

- [x] Successful prevention of none selected
- [x] Successful prevention of "Don't know" + another race
- [x] Allows single or multiple races
- [ ] ETHISPANIC is correctly grouped with racial group specifications
- [ ] Allows only race and does not require ethnicity
- [ ] Allows any combination of ethnicities even if race is unknown

PR #475, opened 2025-07-10
Updates

### LINKED ISSUE #458: D1a IMPNOMCI currently handled as a bool in the export
### IMPNOMCI incorrectly handled as boolean in the export
- currently, our export handles booleans and integers differently. NACC expects all checkboxes to be 1 or null if is a checkbox on the form and 1, 0, or null if it is a radio button. For the export to identify which of these are radio buttons and which are checkboxes so we can handle these in the export, all booleans are assumed to be checkboxes where a false value (0) is nulled out to the export. 

### Steps to reproduce
1. Select value 0 for IMPNOMCI in the form D1a
2. Finalize this form and prepare for submission
3. Download the export for this packet
4. The value for IMPNOMCI in the export will be blank instead of "0"

### Resolution
- [ ] Change the formfield for IMPNOMCI to be a `int?` from a `bool?` in the formfields and adjust the mappings to match 

#### IMPNOMCI is a bool? type
Image

#### The value for this field is 0
Image

#### as a result, the value in export is assumed to be null since it is handling it as a checkbox value

Image

ahmedhaly7777:
Fixes #458 

IMPNOMCI

### DIFF
--- a/src/UDS.Net.Forms/Extensions/DomainToViewModelMapper.cs
+++ b/src/UDS.Net.Forms/Extensions/DomainToViewModelMapper.cs
@@ -1405,7 +1405,7 @@ namespace UDS.Net.Forms.Extensions
                 IMPNOMCLCD = fields.IMPNOMCLCD,
                 IMPNOMCIO = fields.IMPNOMCIO,
                 IMPNOMCIOX = fields.IMPNOMCIOX,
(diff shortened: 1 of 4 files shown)


ashleybot (approved):
@ahmedhaly7777 Looks good! Good to bump patch version, squash merge to main, and tag that commit to build pkgs.

Tested:

- [ ] When IMPNOMCI is null then nothing is exported
- [ ] When IMPMNOMCI has a value, specifically 0, then 0 is exported

Screenshot 2025-07-14 at 8 09 25 AM

Issue #276, opened 2025-02-12
Clearing RequiredIfRange error messages after parent value is modified to within range


ashleybot:
When an error message is displayed for a field and the user updates the field with a value within the valid range, the error message should disappear.

When the reason for the error message to be displayed has been resolved, the error message on an input should react and remove the error message and error message styling from view

Examples:
- When a child is required when parent is within range, and then parent is set outside of range. The error message should not display if child is no longer required
- When disabled state is set for input, the error styling is overriding the disable style. In the screenshot, when 13a is set for 95, the children should disable and the styling should revert to disabled style

![Image](https://github.com/user-attachments/assets/6b03affe-2944-480f-9e99-d9178e944486)

#### aspnet core documentation / jquery unobtrusive JS: https://github.com/dotnet/aspnetcore/blob/ed1b91f5aa6d87194eb9b18a4b1d526e5a97c95d/src/Security/samples/Cookies/wwwroot/lib/jquery-validation-unobtrusive/jquery.validate.unobtrusive.js

PR #482, opened 2025-07-16
Required on range clear

mlan225:
Resolves: #276
Fixes: #489


mlan225:
### Progress
Testing removing validation error styles and text in children when watched value is out of bounds with section 13 of the C2 

#### After submitting with errors
image

#### After adjusting watched value (REYDREC) to out of bounds value (95)
image

mlan225:
Going to impliment the `resetForm()` method for clearing input styles. Does not seem to clear error messages from disabled children

src/UDS.Net.Forms/wwwroot/js/jquery.unobtrusive.custom_validation.js
```diff
@@ -285,214 +288,234 @@ $.validator.addMethod("birthmonth", function (value, element, params) {
...
+                    let lowInt = parseInt(options.params.lowvalue);
+                    let highInt = parseInt(options.params.highvalue);
+                    if (watchedInt <= lowInt || watchedInt >= highInt) {
+                        // outside the range, reset everything
+                        let element = $(options.element);
+                        if (element.length) {
+                            // reset css
+                            element.removeClass("input-validation-error");
+                            element.addClass(
+                                "block w-full max-w-lg rounded-md border-gray-400 shadow-sm focus:border-indigo-500 focus:ring-indigo-500 sm:max-w-xs sm:text-sm placeholder:text-gray-400 disabled:bg-slate-50 disabled:text-slate-500 disabled:border-slate-200 disabled:shadow-none",
```
ashleybot, line 487: @mlan225 Generally, adding hard-coded css will eventually break something and be difficult to investigate. Unobtrusive JS shouldn't remove existing classes, so you shouldn't need to add them again. That said, unobtrusive JS does add a `valid` class. What was the issue that you were trying to solve here?

ashleybot (changes requested):
@mlan225 This is shaping up! The validation summary error messages should also be cleared.

### RequiredIfRange correctly requiring and displaying error messages in proximity to input
Screenshot 2025-07-31 at 8 21 12 AM

### Error messages also displayed in validation summary in form footer
Screenshot 2025-07-31 at 8 21 19 AM

### Error messages are correctly cleared when field should no longer be required
Screenshot 2025-07-31 at 8 21 42 AM

### Error messages in summary are _incorrectly_ displayed after valid state reached
Screenshot 2025-07-31 at 8 21 50 AM

mlan225:
### Progress update
- [ ] Refactor and clean up JavaScript custom_validation method and adapter code
- [ ] Resolve issue with validation between a combination of radio and input parent elements
- [ ] Resolve any bugs that may have arrived when unobtrusive validation was now being used with radio inputs
  - mainly style issues now that classes were being applied to radio buttons, styling was set for only number inputs
- [ ] Error messages are still being removed requiredifrange validation child elements are disabled by parent
- [ ] Error messages are being removed from validation summary when requiredifrange child validation elements are disabled by parent
- [ ] Error messages on element are still being removed when requiredIfrange child element is given a value after unobtrusive validation runs on attempted submission

### Remaining tasks
- [ ] Now that requiredifrange elements are properly activated by unobtrusive validation, elements in the A5D2 (there could be more on other forms) are clashing if they have both the requiredifrange AND requiredif custom attributes. Error detected on these types of elements when they should be valid 
- [ ] When an element is given a value after an error submission caught by unobtrusive, the code that removes errors next to the field is not handled by the requiredifrange adapter or the method, so the validation summary error is not removed in this instance, as it is not default functionality from unobtrusive.

mlan225:
### Required if regular expression
- Spots that use both requiredif and requiredifrange should use the new requiredifregularexpression helper

mlan225:
.NET regular expression attribute: https://github.com/microsoft/referencesource/blob/main/System.ComponentModel.DataAnnotations/DataAnnotations/RegularExpressionAttribute.cs

mlan225:
### Progress update
- [ ] Impliment the requiredifregex attribute
- [ ] Implement the requiredifregex unobtrusive validation logic
- [ ] Verify that all instances of both requiredif and requiredifrange are replaced with requiredifregex

### Final steps
- [ ] The unobtrusive main library still has the custom override that adds IDs to the error message summary. Find out if there is a way to override this functionality so we can persist it on update.
  - The slight issue with this is that I would need to find a way to override a specific function of the base library code, so the rest of the code can be updated by nuget if anything changes. It may be worth it to just leave the error message summary as is 🤔, overriding any part of that code could have a negative effect on updates to the main library file

mlan225:
#### Unobtrusive validation feature breakdown:

[Unobtrusive validation.pdf](https://github.com/user-attachments/files/22166798/Unobtrusive.validation.pdf)

mlan225:
Unobtrusive validation PR powerpoint

[Unobtrusive Validation.pptx](https://github.com/user-attachments/files/22329616/Unobtrusive.Validation.pptx)

src/UDS.Net.Forms/Models/UDS4/A5D2.cs
```diff
@@ -460,7 +460,7 @@ public class A5D2 : FormModel
         [Display(Name = "Choose if treated or untreated")]
         [RegularExpression("^([0-1])$", ErrorMessage = "Valid range is 0-1")]
         [RequiredIf(nameof(OTHERDEP), "1", ErrorMessage = "Response required")]
-        [RequiredIf(nameof(MAJORDEP), "1", ErrorMessage = "Response required")]
+        [RequiredIfRegex(nameof(MAJORDEP), "1", ErrorMessage = "Response required")]
```
ashleybot, line 463: @mlan225 I'm not sure if this one should be changed.

src/UDS.Net.Forms/Models/UDS4/D1b.cs
```diff
@@ -131,7 +122,7 @@ public bool? SPECTValidation
         [Display(Name = "Elevated tau pathology")]
         public int? TAUPET { get; set; }
 
-        [RequiredIfRange(nameof(PETDX), 1, 2, ErrorMessage = "Please specify.")]
+        [RequiredIfRegex(nameof(IMAGINGDX), "(1|3)", ErrorMessage = "Please specify.")]
```
ashleybot, line 125: @mlan225 This validation for `FDGPETDX` should not be changed.

src/UDS.Net.Forms/Pages/UDS4/D1b.cshtml
```diff
@@ -162,7 +162,6 @@
                     <span class="mt-2 text-sm text-red-600" asp-validation-for="D1b.CSFAD"></span>
                 </td>
                 <table-cell-radio-buttons id="@Html.IdFor(m => m.D1b.CSFAD)" for="@Model.D1b.CSFAD" items="Model.FindingsListItems"></table-cell-radio-buttons>
-            </tr>
```
ashleybot, line 165: @mlan225 This closing tag should not be removed.

ashleybot (changes requested):
@mlan225 Nice work! I added a commit to fix Playwright tests. It looks like there are a few copy-paste errors. Please see comments.

src/UDS.Net.Forms/Models/UDS4/A5D2.cs
mlan225: Good to keep change

src/UDS.Net.Forms/Models/UDS4/D1b.cs
mlan225: Good to keep change

mlan225:
### Updates from review
- [ ] Added back in removed table row end element
- [ ] Double checked production data to make sure the attribute validation change for `FDGPETDX` did not require any database changes. Did not appear to need any.

ashleybot (approved):
@mlan225 Let's bump the major version with this one. It isn't technically a breaking change, but it is a big milestone! Nice work!

Issue #486, opened 2025-07-24
C2 MINTSCNC validation enhancement


mlan225:
# MINTSCNC validation and further NACC questions

Josh received these errors in section 16 from a submission in the C2 form:

The values included in the submission error

<img width="1201" height="560" alt="Image" src="https://github.com/user-attachments/assets/a7a80ff4-3a19-40d6-8a89-d2f58e80aa35" />

## MINTSCNC validation rule
> MINTSCNC validation rule:
if q16c. mintscng (mint number semantic cues given) is > 0 then q16d. mintscnc (mint correct with semantic cue) must be less than or equal to q16c. mintscng (mint number semantic cues given)

NACC is validating MINTSCNC in the submission error data as failed because 88 is not counted as an "other" variable, and is larger than MINTSCNG 

### Steps to reproduce error
1. Input a value of 1 for MINTSCNG
2. Input a value of 88 for MINTSCNC
3. Finalizing form will NOT produce an error with MINTSCNC when there should be one

## MINTTOTS validation rule
> MINTTOTS validation rule:
if q16a. minttots (mint total) is between 0 and 32 and q16d. mintscnc (mint correct with semantic cue) =88, then q16a. minttots must = q16b. minttotw

NACC is requiring MINTTOTS to equal MINTTOTW when MINTSCNC is 88

### Steps to reproduce
1. Input a value of 32 for MINTTOTS
2. input a value of 29 for MINTTOTW
3. Input a value of 88 for MINTSCNC
4. Finalizing form will NOT produce an error when MINTTOTS does not equal MINTTOTW when there should be one

## To do
- [ ] Implement the MINTSCNC validation rule
- [ ] Implement the MINTTOTS validation rule

mlan225:
There is an issue I spoke about with Josh that may be worth discussing further

### Ambiguity for MINTSCNC and MINTPCNC values

<img width="990" height="300" alt="Image" src="https://github.com/user-attachments/assets/68945f06-7588-42ed-8816-4b270ebc340a" />


### MINTSCNC ambiguity 
Right now, NACC validation rules are expecting 16d to be either 88 or 0 when 16c is 0. RIght now there is no reference in the NACC documentation as to when 16d should be 0 or 88, and the data shows these values used interchangeably in our database.

The issue here is that putting 88 activates further validation and forces values, so there are implications for putting 88 that 0 does not have.
[...]

PR #491, opened 2025-07-30
mint validation enhancement

mlan225:
Resolves: #486 

Draft PR

### DIFF
--- a/src/UDS.Net.Forms/Models/UDS4/C2.cs
+++ b/src/UDS.Net.Forms/Models/UDS4/C2.cs
@@ -338,9 +338,9 @@ namespace UDS.Net.Forms.Models.UDS4
             {
                 if (RMMODE != RemoteModality.Telephone)
                 {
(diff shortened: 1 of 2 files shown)


src/UDS.Net.Forms/Models/UDS4/C2.cs
```diff
@@ -336,11 +408,11 @@ public bool? MINTSCNCValidation
         {
             get
             {
-                if (RMMODE != RemoteModality.Telephone)
+                if (MODE == FormMode.InPerson || (MODE == FormMode.Remote && RMMODE == RemoteModality.Video))
```
mlan225, line 411: Adjusted condition to be more accurate than "not telephone"

ashleybot (approved):
@mlan225 I added a few tests to handle some MINT scenarios. All pass! 

I think there are some UX improvements that can be made with more readable and friendly error messages, but this doesn't need to hold up the card. 

Good to bump patch version and merge to main, or if you want to edit the messages, just put it back into review when you're ready again.

Issue #377, opened 2025-04-04
B5 A5D2 Cross-form validation - Anxiety


ashleybot:
Returned alert from NACC submission

> if q6a. anx (anxiety) = 0 (no), then form a5d2, q6d. anxiety (anxiety disorder) should not equal 1 (active)

Add validation check on visit finalize.

**B5**
![Image](https://github.com/user-attachments/assets/b4d45225-63fd-4c86-92ec-b33e5d788a22)

**A5D2**

![Image](https://github.com/user-attachments/assets/2cf44ed4-9d9e-4729-a78d-585c3f7ab544)

PR #495, opened 2025-08-05
Search Milestones Index

### LINKED ISSUE #337: Search Milestones Index
This card includes all changes to UI, razor page, service layer and interfaces.

- [x] Add ability to search milestones index by Legacy ID
- [x] Utilize existing standards for search UI (see participants index)
- [ ] Use Turbo Frames
- [x] Add method to IMilestoneService to FindByLegacyId
- [x] Implement method in UDS.Net.Web.MVC.Services.MilestoneService to call underlying api client method GetMilestonesByLegacyIdAndStatus

NOTE: You should be able to use MilestoneClient.GetMilestonesByLegacyIdAndStatus(string legacyId, string[] statuses, int pageSize = 10, int pageIndex = 1), but it may need to be adjusted. Please discuss if needed.

NOTE: This will create a breaking change so major version number will be increased. Corresponding feature branch in local repo should be created and tested using preview packages.

Oddvocado:
Resolves #337

### DIFF
--- a/src/UDS.Net.Forms/Pages/Milestones/Index.cshtml
+++ b/src/UDS.Net.Forms/Pages/Milestones/Index.cshtml
@@ -7,7 +7,36 @@
 <div class="space-y-16 py-16 xl:space-y-20 px-4 sm:px-6 lg:px-8">
 
     @if (Model.Milestones != null && Model.Milestones.List != null)
+
     {
+        <div class="relative flex justify-between lg:gap-8 xl:grid xl:grid-cols-12">
+            <div class="flex md:absolute md:inset-y-0 md:left-0 lg:static xl:col-span-4">
+                <div class="px-4 sm:px-6 lg:px-8">
+                    <h2 class="mx-auto max-w-2xl text-base font-semibold leading-6 text-gray-900 lg:mx-0 lg:max-w-none">All milestones</h2>
+                    <p class="mt-2 text-sm text-gray-700">A list of all milestones.</p>
+                </div>
+            </div>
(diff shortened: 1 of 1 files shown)


Oddvocado:
Quick sketch of the UI. This will mimic the existing UI in the Participation endpoint of the app to maintain a consistent look and feel.

#377 UI Sketch

Oddvocado:
@ashleybot this is ready for review!

src/UDS.Net.Web.MVC.Services/MilestoneService.cs
```diff
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Threading.Tasks;
+using System.Xml;
```
ashleybot, line 4: @Oddvocado Shouldn't need the Xml library here.

ashleybot (changes requested):
@Oddvocado Can you remove the word "legacy" from the placeholder text to make it more agnostic as to which ID? We use NACC ID for other centers and PT ID for our own, so it will make things less confusing for the user.

Image

src/UDS.Net.Web.MVC.Services/MilestoneService.cs
Oddvocado: This one snuck past me! I've gone ahead and removed it, I've also updated the placeholder text in the milestone search bar.

Oddvocado:
@ashleybot ready for another review!

ashleybot (approved):
@Oddvocado Looks good! This is going to be a major version bump because it will introduce breaking changes. Good to merge to main and open a PR with updates in our local repo.

PR #506, opened 2025-08-20
Update Milestone Service 

Oddvocado:
This PR adds the username parameter to the FindByLegacyId service so that the COA.Uds.v4 application can implement the service.


ashleybot (approved):
@Oddvocado Update to uds-dotnet-services looks good!

PR #531, opened 2025-09-10
Fixed validation on REFERSCX

### LINKED ISSUE #527: A1 REFERSCX should be required if REFERSC == 8
**Describe the bug**
REFERSCX allows null when REFERSC == 8, but it should require a value.

**To Reproduce**
1. Start a new or existing A1
2. Fill out all values required for A1 to be finalized
3. Select Option 8 Other for Question 25 REFERSC
4. Do not enter a value for REFERSCX
5. Save Status to Finalized
6. Click "Save"
7. Form incorrectly finalizes and redirect to visit details

**Expected behavior**
If REFERSC == 8 then REFERSCX should require a value and the form should not allow finalization.

**Screenshots**

Image

CallowaySimon:
Resolves #527


ashleybot (approved):
@CallowaySimon Great work! Thanks for getting the fix in to review so quickly. Good to bump to stable patch and merge to main.

PR #542, opened 2025-09-16
Adjust the formkind match to compare both with ToUpper()

mlan225:
Resolves: #557 
The resolution to this was within the `uniform-data-set-dotnet-web` project

## Fixed the null assignedTo and min dateTime CreatedAt value in packetsubmission error
- The packet submission error was giving NULL to some packet submission errors as well createdAt being mininum datetime

#### Adjust comparison logic
-  Form data was being selected based on a matching uppercase formkind string. NACC returns form kinds as lower case and our data has formkinda with with a mix of upper case and lower case. Originally the code was written to make the NACC formkind uppercase, but that lead to formkinds in our database with a combination of uppercase and lowercase to not be found (D1a, A1a, etc.).
This also resulted in a min date createdAt because a record was not found

image

## Additional resources
Test error file for upload: [table_data_form-qc-checker (85)(in).csv](https://github.com/user-attachments/files/22370427/table_data_form-qc-checker.85.in.csv)
- you will need to adjust the `ptid` and `visitnum` for each error row to match your packet submission. 
- The last numbers in the `code` column do not need to be anything specific, as those are not needed to submit with a packet, but is data returned from NACC.

## Checklist
- [ ] Formkinds in packetSubmissionError with 3 letters are now saved with an assignedTo and correct CreatedAt datetime value
- [ ] Existing 2 letter formkinds in packetSubmissionError are still saved with proper assignedTo and CreatedAt values
- [ ] The Unique formkind `A5D2` in packetSubmissionError is still saved with proper assignedTo and CreatedAt values
  - This formkind is the only one with more than 2 letters and all uppercase

##### (Personally tested screenshot from PR, this the correct result  for reference)
image

### DIFF
--- a/src/UDS.Net.Forms.Tests.Runtime/UDS.Net.Forms.Tests.Runtime.csproj
+++ b/src/UDS.Net.Forms.Tests.Runtime/UDS.Net.Forms.Tests.Runtime.csproj
@@ -4,6 +4,7 @@
 		<TargetFramework>net8.0</TargetFramework>
 		<Nullable>enable</Nullable>
 		<ImplicitUsings>enable</ImplicitUsings>
+		<ReleaseVersion>13.0.22</ReleaseVersion>
 	</PropertyGroup>
 	<PropertyGroup Condition=" '$(RunConfiguration)' == 'https' " />
 	<PropertyGroup Condition=" '$(RunConfiguration)' == 'http' " />
--- a/src/UDS.Net.Forms.Tests/UDS.Net.Forms.Tests.csproj
+++ b/src/UDS.Net.Forms.Tests/UDS.Net.Forms.Tests.csproj
@@ -5,7 +5,7 @@
 		<ImplicitUsings>enable</ImplicitUsings>
 		<Nullable>enable</Nullable>
 		<IsPackable>false</IsPackable>
-		<ReleaseVersion>13.0.15</ReleaseVersion>
+		<ReleaseVersion>13.0.22</ReleaseVersion>
 	</PropertyGroup>
 	<ItemGroup>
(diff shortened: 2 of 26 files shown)


src/UDS.Net.Forms/Pages/PacketSubmissionErrors/Create.cshtml.cs
```diff
@@ -75,10 +75,8 @@ public async Task<IActionResult> OnPostAsync()
                     formAssignee = formAssignedTo;
                 }
 
-                DateTime createdAt = formData.Select(a => a.CreatedAt).FirstOrDefault();
-
                 //Create new PacketSubmissionError object with collected data
-                PacketSubmissionError newPacketSubmissionError = new PacketSubmissionError(0, PacketSubmissionId, formKind, error.Message, formAssignee, errorLevel, User.Identity.Name, createdAt, User.Identity.Name, null, null, false);
+                PacketSubmissionError newPacketSubmissionError = new PacketSubmissionError(0, PacketSubmissionId, formKind, error.Message, formAssignee, errorLevel, User.Identity.Name, DateTime.Now, User.Identity.Name, null, null, false);
```
mlan225, line 79: My previous confusion came from the now deleted line: 

```
DateTime createdAt = formData.Select(a => a.CreatedAt).FirstOrDefault();
```

The `formData` being set to the form found in a packet, so looking at the code `CreatedAt` looks like it *should* have been the createdAt date for that form in the packet. However, the created at value was being set to `DateTime.Now` in the form domain model. So that was correct.

To make this more explicit and avoid confusion in the future, I added dateTime.Now to the constructor call instead of getting it from packet data.

PR #545, opened 2025-09-22
Form footer mode, rmreas, rmmode, not

### LINKED ISSUE #540: Form Footer MODE, RMREAS, RMMODE, NOT
**User can select a value when they should not be able to**
RMREAS, RMMODE, NOT states are not loaded when the page loads (they are correctly set when MODE changes). So, the user is able to select and persist a RMREAS, RMMODE, and/or NOT value when they should not be able to.

**To Reproduce**
1. Go to a form such as the A2
2. Scroll down to the bottom
3. See `MODE` is set to InPerson (If MODE is InPerson the user should not be able to set RMREAS, RMMODE, or NOT)
4. Choose 'Telephone' for Remote modality `RMMODE`
5. Choose 'Other' for Not completed reason `NOT`
6. Save the form
7. See the data is incorrectly persisted in the db

Image

**Expected behavior**
When the page loads, the UI state of RMREAS, RMMODE, and NOT should behave just the same as when MODE is changed.

Image

CallowaySimon:
Resolves #540 

## MODE, RMREAS, RMMODE, NOT

Disable RMREAS, RMMODE, NOT on page load when in person is selected.

image

### Review Checklist

- [x] RMREAS, RMMODE, NOT are disabled on page load when in person in selected
- [x] RMREAS, RMMODE, NOT are not cleared on page load when another value has been saved for MODE

### DIFF
--- a/src/UDS.Net.Forms.Tests.Runtime/UDS.Net.Forms.Tests.Runtime.csproj
+++ b/src/UDS.Net.Forms.Tests.Runtime/UDS.Net.Forms.Tests.Runtime.csproj
@@ -4,7 +4,7 @@
 		<TargetFramework>net8.0</TargetFramework>
 		<Nullable>enable</Nullable>
 		<ImplicitUsings>enable</ImplicitUsings>
-		<ReleaseVersion>13.0.19</ReleaseVersion>
+		<ReleaseVersion>13.0.20</ReleaseVersion>
 	</PropertyGroup>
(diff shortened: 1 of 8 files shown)


ashleybot (approved):
@CallowaySimon Works great! I loaded several forms, new and existing, and the footer functioned as expected. There was one error on the C2, but that fix is pending and should be resolved with #482. Good to bump to next stable version and merge to main.

Issue #536, opened 2025-09-16
C2 UDSVERTI Validation


ashleybot:
Add validation for Question 11i `UDSVERTI`. When `UDSVERNF` and `UDSVERLN` are both within valid ranges (0-15) then `UDSVERTI` should equal the sum.

<img width="1261" height="1200" alt="Image" src="https://github.com/user-attachments/assets/dcb9b13c-5368-4e51-9b80-c16daeba87da" />

### Referenced error
> if q11c. udsvernf (f-words rule violations) is between 0 and 15 and q11f. udsverln (l-words rule violations) is between 0 and 15, then q11i. udsverti must be the total of q11c. udsvernf and q11f. Udsverln

PR #547, opened 2025-09-22
Add UDSVERTI validation

mlan225:
Resolves: #536 

### Add validation for UDSVERTI
Add validation for Question 11i UDSVERTI. When UDSVERNF and UDSVERLN are both within valid ranges (0-15) then UDSVERTI should equal the sum.

**UDSVERNF, UDSVERLN, and UDSVERTI questions are on BOTH the in-person and telephone forms**

### Task List (apply for both in-person and telephone views)
- [ ] If UDSVERNF and UDSVERLN have values of 0 - 15, then UDSVERTI must be the sum of UDSVERNF + UDSVERLN
- [ ] If either UDSVERNF and UDSVERLN do not have a value, then validation does not run
- [ ] Validation error message appears on the right-hand column next to the input, following the established validation message location for the section
- [ ] Form saves properly when validation is satisfied

### DIFF
--- a/src/UDS.Net.Forms/Models/UDS4/C2.cs
+++ b/src/UDS.Net.Forms/Models/UDS4/C2.cs
@@ -535,6 +535,26 @@ namespace UDS.Net.Forms.Models.UDS4
         [RequiredIfRange(nameof(UDSVERLC), 0, 40, ErrorMessage = "Provide count of non-F/L words and rule violation errors.")]
         public int? UDSVERTI { get; set; }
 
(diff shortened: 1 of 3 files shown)


mlan225:
### Review changes
- Adjusted validation attribute for `UDSVERTIValidation` to check for `UDSVERLC` to be 0 - 40 before running
- Slightly adjusted the HTML to display error messages on UDSVERTI on separate lines if the value was both missing and required a total sum value 
  - Separated validation messages with div elements to stack them instead of having them on the same line

Issue #550, opened 2025-09-23
Packet submission Success and view visit bugs


mlan225:
### Packet submission and View visit error in the packet submissions view
- Selecting "success" on a packet submission results in an error, but still sets the success status

- Selecting "view visit" on the packet submission page results in an error 

<img width="1831" height="744" alt="Image" src="https://github.com/user-attachments/assets/28c41bcd-e423-480c-8fc0-8beacc1a8343" />

PR #551, opened 2025-09-23
Address packet submission bugs

mlan225:
Resolves: #550 

This PR tackles 3 different tasks
- Address the packet submission "success" link bug
  - clicking the success link caused an error due to the `packet.errors` object being NULL when no errors are submitted. The code now has a NULL check.
- Address the packet submission "View visit" button bug
  - The view visit button was causing an error due to the partial providing participation id rather than the visit id
- Move the "view visit" button in packet submissions 
  - The "view visit" button is now in line with the "+ new packet submission" button

### Packet submission success
- [ ] When viewing a submitted packet in the packet submission view, selecting the "success" link will no longer result in an error
### Packet submission View visit
- [ ] The "view visit" will not result in an error on the uds.web public application
- [ ] The "view visit" will no longer result in an error on the uds4 - local application
  - Running the package version `13.0.20-preview.1` will allow you to test the uds4 - local application
### Reposition the "view visit" button 
- [ ] "View visit" button on packet submissions is now in line with the "+ new packet submission" button

### DIFF
--- a/src/UDS.Net.Forms.Tests.Runtime/UDS.Net.Forms.Tests.Runtime.csproj
+++ b/src/UDS.Net.Forms.Tests.Runtime/UDS.Net.Forms.Tests.Runtime.csproj
@@ -4,7 +4,7 @@
 		<TargetFramework>net8.0</TargetFramework>
 		<Nullable>enable</Nullable>
 		<ImplicitUsings>enable</ImplicitUsings>
-		<ReleaseVersion>13.0.23</ReleaseVersion>
+		<ReleaseVersion>13.0.24</ReleaseVersion>
 	</PropertyGroup>
 	<PropertyGroup Condition=" '$(RunConfiguration)' == 'https' " />
 	<PropertyGroup Condition=" '$(RunConfiguration)' == 'http' " />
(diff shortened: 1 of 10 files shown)


mlan225:
## View visit bug progress

Was able to confirm the root issue as to why the links were working on the public app, but not our local (uds4) 

I originally thought the local application was grabbing the id for the route differently, but I was able to confirm that the route does not work correctly in both the public and local applications. The reason it works in the public application is because, with local data its a common scenario for this order of events:
- Create participation
- Create visit for that participation
- Fill out forms 
- Submit packet

This pathway causes a coincidence where participationId and visitId are the same, thus, allowing the "View visit" button to work with particpationId because it just so happened to be the same as visit Id. 


### Resolution
- Adjust the "view visit" button to use the visitId instead of participation Id. This should fix it on both applications.

mlan225:
### Buttons updated to be inline
image

mlan225:
Encountered an error when testing the preview version with uds4 - local

When viewing a visit, the uds4 - local application seems to encounter an error with viewing the `_FormsGridPartial`.

uds4 main seems to be running version `13.0.17`, which would be 2 versions behind the current version `13.0.19`. So I wonder if this happened in the last 2 versions.

I will look into the resolution and add it to the PR

ashleybot (approved):
@mlan225 Looks good! Good to bump to next stable patch and merge to main.

PR #572, opened 2025-10-22
Update PacketSubmission ToDomainMapper

CallowaySimon:
Update the PacketSubmission ToDomain mapper so that Forms and Errors can be mapped.

Also updated PacketSubmission Domain Model constructors to support changes


ashleybot (approved):
@CallowaySimon Looks good!

PR #577, opened 2025-10-23
Disable CDOMBEH when MCI = 1

### LINKED ISSUE #568: D1a Question 6f CDOMBEH should be disabled when MCI = 1
There are specific scenarios when Question 6f should be enabled.

1. We are already handling the first scenario when NORMCOG = 1, the checkboxes are correctly disabled.
2. We are already handling the second scenario when MCI = 0, the checkboxes are correctly disabled.
3. In the third scenario, follow the path when DEMENTED = 0 and MCI =1, the checkboxes are all enabled, but CDOMBEH should be disabled.

In other words, when the Question 6 checkboxes are enabled because DEMENTED = 1, then they should all be enabled and selectable. When Question 6 checkboxes are enabled because MCI = 1, only a portion of the checkboxes should be enabled; CDOMBEH should be disabled.

### DEMENTED = 1

Image

Image

### DEMENTED = 0, MCI = 1

Image

Image

Image

> Form D1a: In “Affected Domains – Dementia and MCI section instruction (above Q6)”: Change “For patients
> with dementia who have behavior changes, record the presence of behavioral changes here (not in the following MBI
> section) by marking Q6f as Impaired and skipping the MBI section (SKIP TO Q8).” to “For patients with dementia who
> have behavior changes, record the presence of behavioral changes here (not in the following MBI section) by marking Q6f as Impaired and Q7 as 0 = No.”

mlan225:
Fixes #568 

## Validation enhancement for D1a CDOMBEH
When MCI = 1, then question 6f CDOMBEH should be disabled. When DEMENTED = 1, then all checkboxes should be enabled

image

### New validation
- [x] Disable only 6f CDOMBEH when MCI = 1

### Existing validation that should remain intact
- [x] CDOMBEH and all other checkboxes in question 6 should be enabled when DEMENTED = 1
- [x] All checkboxes in question 6 should be disabled when MCI = 0
- [x] When NORMCOG = 1, question 6 checkboxes are disabled

### DIFF
--- a/src/UDS.Net.Forms.Tests.Runtime/UDS.Net.Forms.Tests.Runtime.csproj
+++ b/src/UDS.Net.Forms.Tests.Runtime/UDS.Net.Forms.Tests.Runtime.csproj
@@ -4,7 +4,7 @@
 		<TargetFramework>net8.0</TargetFramework>
 		<Nullable>enable</Nullable>
 		<ImplicitUsings>enable</ImplicitUsings>
-		<ReleaseVersion>15.0.5</ReleaseVersion>
+		<ReleaseVersion>15.0.6</ReleaseVersion>
 	</PropertyGroup>
 	<PropertyGroup Condition=" '$(RunConfiguration)' == 'https' " />
 	<PropertyGroup Condition=" '$(RunConfiguration)' == 'http' " />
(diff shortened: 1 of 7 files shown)


ashleybot (approved):
@mlan225 Changes look good. Good to bump to next stable patch and merge to main. 

I added a commit to update the Playwrights tests. If there are existing tests for a form that you are working on, part of your dev process should be to update the Playwright tests along with the feature if it makes sense. In this case, because MCI = 1 was checking for an **enabled** CDOMBEH which is incorrect and the tests would have failed if you ran the tests locally.

PR #583, opened 2025-10-27
Limit packet submissions

### LINKED ISSUE #310: Only allow new packet submissions if visit status is in specific states
Confirm that the following is true and add end-to-end tests.

## Allow new packet submission if status is
- Finalized

## Do not allow new packet submission if status is

- Pending
- Submitted (pending results of submission)
- FailedErrorChecks
- PassedErrorChecks
- Frozen

## Reference

- [Packet status](https://github.com/UK-SBCoA/uniform-data-set-dotnet-web/blob/main/src/UDS.Net.Services/Enums/PacketStatus.cs)

CallowaySimon:
Resolves #310 

## Limit packet submissions

Add validation when attempting to submit a packet to ensure its status is finalized. 

image


ashleybot (approved):
@CallowaySimon Thanks for thinking through this and talking me through things yesterday. Good to bump to next stable patch and merge to main.

I added a small commit to display the packet status on the details view.

PR #584, opened 2025-10-28
C2 and C2T REYDREC Validation

mlan225:
Resolves: #620
Resolves: #621

## Rework the C2 form
After some discussion, we have decided to rework the C2 form from using JavaScript to switch partial views to using form submissions to return the partials to resolve playwright test bugs. This change led to a full rework of the C2 form background functionality while also resolving some of the original issues that began this PR.

The review checklist below will involve changes/fixes to the C2 partial views as well as the new form switch functionality.

### C2 Form view switch functionality
Previously, we javascript and Turbo to switch form views for the C2. Now, we are using form submissions within the partial views to request partial views from the server and have them rendered into the turbo frame on the C2.cshtml. 

C2 form submission flow

### C2 layout changes
The layouts have been adjusted so the C2 can provide its own necessary hidden fields directly and other forms without a special layout to use the `_LayoutForm` partial to include these fields

This change is because the C2 partials need to submit these hidden fields inside of their respective _C2 and _C2T form elements manually

C2 layout changes

### Validation adjustments
REY1REC and section 13 disable/enable functionality has been adjusted. REY1REC no longer disables 13b to the end of the section. These question are now controlled ONLY by REYDREC (first question in section 13)

> The before and after screenshots will refer to the InPerson view of the C2. This logic applies to both views.

|Before| After|
|-------|-------|
|Applying 95-98 to REY1REC results in disable of section 13 questions. When section 13 has values previously, they are cleared (aside from REYDREC) and upon save trigger a validation error for all the cleared inputs| Applying 95-98 to REY1REC no longer controls disable state of section 13 questions. As a result, when section 13 has values previously, only the server side validation is triggered because the previous values are NOT cleared (for all except REYDREC)|
|image| image |
|image| image|
| image| image |

### Test adjustments to VisitTests
Adjustments made to the test file `VisitTests/CheckForForms`. These changes seem to have resolved previous issues with these tests when running in GitHub actions.

Small adjustments were made to add additional info or focus for test lines, however, I will outline the primary changes to `VisitTests`

#### Before
```
1. get  by role with Name = "Go"

2. `WaitForURLAsync("**/Visits/Details/*")`

3. Assert.IsTrue for all necessary forms
```

#### After
```
1. get  by role with Name = "New visit"

2. `WaitForRequestFinishdAsync()`

3. Expect all necessary form elements to be visible
```

### view adjustments
- [x] Display property names back to the _C2T (Telephone form) section 6 UI labels
- [x] C2 and C2T partials have been adjusted to use 2 forms, one for switching views and one for posting form data on save
- [x] Hidden fields have been added to C2 and C2T partials to allow form submission from within the partial 
- [x] `C2.cshtml` has been adjusted to use its own C2 layout 
  - The original C2 layout included hidden fields relevant to form submission. Since we are now posting forms from within the partials, we needed to include these fields in the partials themselves
- [x] Layouts were adjusted to remove code that specifically handled the C2 form and C2 layouts were created in place of this
  - The C2-specific code now has its own partials to handle C2 scenarios and does not need to be accounted for in the base layouts for all forms

### File removals and additions
- [x] `C2.js` was removed
- [x] `c2_controller.js` was added to handle C2 javascript
- [x] `selectendpoint.js` file was removed
  - This file was used specifically to handle the view switching in the C2. This process is now handled by form posts
- [x] Remove legacy file `_C2Validation.cshtml` partial
  - This seems to have had some use for the previous turbo + javascript form switch functionality 

### Form functionality
- [x] In the C2 form on the C2T view, validation will now stop you from finalizing when question 6a. is 95-98 and question 13a is not 95 - 98
- [x] Switching between in-person and remote telephone partials (_C2 and _C2T) will now reapply the disabled states from client-side validation after the switch
  - Global Javascript method was created on custom validation javascript file so the c2.js can access and call it on view change
- [x] The C2 form now uses a combination of form submissions and turbo frames to return a new view after selecting the relevant mode and modality options
  - This change was to allow postback from the server on view switch, allowing Playwright tests to successfully await page loads and remove race condition bugs

### Playwright Tests
- [x] Tests added to verify that the REYDREC value needs to be a value of 95 - 98 when REY1DREC is 95 - 98 in both C2 and C2T forms
- [x] Tests added to verify that questions in the REY1DREC section after the first question (12a in C2 | 6a in C2T) are disabled if REY1DREC is a non-complete value (98)
- [x] Tests added to verify that questions in the REYDREC section after 13a (in both the C2 and C2T) are disabled if REYDREC is a non complete value (98) 
- [x] C2MINTTests adjusted to account for if the form was not set for in-person mode on load
  - Test was expecting the mode to already be set for a value before continuing tests. If the form was adjusted by previous test, this would fail. Added logic to test to apply the value it is initially expecting before saving and checking for saved values
- [x] VisitTests adjusted to include missing "A1a" form
- [x] VisitTets/CheckForForms test adjusted to use Expect instead of assert. Removed URL check that was known to fail in headless

### Running tests
- [x] All tests complete successfully in headless mode

### Previous Review Comments
- [x] Removed the unnecessary extra REYDREC validation method in the `C2.cs` file
  - The original intention of this copy was due to the slight change in the error message text, however, this has been adjusted to use wording that points the user to the correct fields regardless of C2 or C2T views
- [x] Removed the duplicate `REY6INT` UIEnableAttribute in the `C2.cshtml.cs`

### More review changes
- [x] Change `_C2Layout` into `_LayoutFormBase` and use the new `_LayoutFormBase` as the base for `_LayoutForm` and `C2.cshtml`
- [x] Remove the turbo frame from `C2.cshtml` to remove double turbo frame loads
- [x] Remove unnecessary code conditionals from `_C2FormFooter` 
- [x] Add disable logic for remote reason when remote mode is selected from the dropdown
  - Noticed this was not done on main and added a fix
- [x] Remove extra validation method for REYDREC
  - Combined both of the C2 and C2T REYDREC validation. Initially, they were separate due to the slight changes in error message text (the question labels were different). I combined them by changing the error message to use the section name instead of the number label on both views

### More review changes (12/08)
- [x] Adjusted the `Requiredif` custom client validation to account for the watched value being a dropdown select option
- [x] Removed dropdown reset data attributes and controller from C2 footer view
  - The C2 needed its own specific js for dropdown mode and related logic
- [x] unnecessary extra C2 check was removed from the FormPageModel layout
  - This section served to provide the related partial, however, the C2.cshtml handles this
- [x] Added code to fix existing bug where the unobtrusive validation did not successfully re-parse the form on form switch
  - The unobtrusive validation loads from the layout, the forms switch with turbo as a child of the layout, so the form needs to be manually reparsed on form switch to allow unobtrusive validation JS to successfully parse the inputs 
- [x] Add missing error messages to C2 properties
  - Without an explicit error message argument for the client-side validation attributes, the error message will not correctly display in the C2

### DIFF
--- a/src/UDS.Net.Forms.Tests.Runtime/UDS.Net.Forms.Tests.Runtime.csproj
+++ b/src/UDS.Net.Forms.Tests.Runtime/UDS.Net.Forms.Tests.Runtime.csproj
@@ -4,7 +4,7 @@
 		<TargetFramework>net8.0</TargetFramework>
 		<Nullable>enable</Nullable>
 		<ImplicitUsings>enable</ImplicitUsings>
-		<ReleaseVersion>15.2.2</ReleaseVersion>
+		<ReleaseVersion>15.3.0</ReleaseVersion>
 	</PropertyGroup>
 	<PropertyGroup Condition=" '$(RunConfiguration)' == 'https' " />
 	<PropertyGroup Condition=" '$(RunConfiguration)' == 'http' " />
--- a/src/UDS.Net.Forms.Tests/UDS.Net.Forms.Tests.csproj
+++ b/src/UDS.Net.Forms.Tests/UDS.Net.Forms.Tests.csproj
@@ -5,7 +5,7 @@
 		<ImplicitUsings>enable</ImplicitUsings>
 		<Nullable>enable</Nullable>
 		<IsPackable>false</IsPackable>
-		<ReleaseVersion>15.2.2</ReleaseVersion>
+		<ReleaseVersion>15.3.0</ReleaseVersion>
 	</PropertyGroup>
 	<ItemGroup>
 		<PackageReference Include="coverlet.collector" Version="6.0.4">
--- a/src/UDS.Net.Forms/UDS.Net.Forms.csproj
+++ b/src/UDS.Net.Forms/UDS.Net.Forms.csproj
@@ -6,13 +6,13 @@
 		<ImplicitUsings>enable</ImplicitUsings>
 		<AddRazorSupportForMvc>true</AddRazorSupportForMvc>
 		<PackageId>UDS.Net.Forms</PackageId>
-		<Version>15.2.2</Version>
+		<Version>15.3.0</Version>
 		<Authors>Sanders-Brown Center on Aging</Authors>
 		<Description>UDS Forms razor class library</Description>
 		<Owners>UK-SBCoA</Owners>
 		<PackageDescription>Razor class library for rendering UDS forms</PackageDescription>
 		<RepositoryUrl>https://github.com/UK-SBCoA/uniform-data-set-dotnet-web</RepositoryUrl>
-		<ReleaseVersion>15.2.2</ReleaseVersion>
+		<ReleaseVersion>15.3.0</ReleaseVersion>
 	</PropertyGroup>
 	<PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' " />
 	<ItemGroup>
--- a/src/UDS.Net.Services.Test/UDS.Net.Services.Test.csproj
+++ b/src/UDS.Net.Services.Test/UDS.Net.Services.Test.csproj
@@ -6,7 +6,7 @@
 		<Nullable>enable</Nullable>
 		<IsPackable>false</IsPackable>
 		<IsTestProject>true</IsTestProject>
-		<ReleaseVersion>15.2.2</ReleaseVersion>
+		<ReleaseVersion>15.3.0</ReleaseVersion>
 	</PropertyGroup>
 	<ItemGroup>
 		<PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.8.0" />
--- a/src/UDS.Net.Services/UDS.Net.Services.csproj
+++ b/src/UDS.Net.Services/UDS.Net.Services.csproj
@@ -4,13 +4,13 @@
 		<TargetFramework>netstandard2.1</TargetFramework>
 		<OutputType>Library</OutputType>
 		<PackageId>UDS.Net.Services</PackageId>
-		<Version>15.2.2</Version>
+		<Version>15.3.0</Version>
 		<Authors>Sanders-Brown Center on Aging</Authors>
 		<Description>Service contracts for implmenting your own back-end with MVC web app</Description>
 		<Owners>UK-SBCoA</Owners>
 		<PackageDescription>Service contracts required by MVC web app</PackageDescription>
 		<RepositoryUrl>https://github.com/UK-SBCoA/uniform-data-set-dotnet-web</RepositoryUrl>
-		<ReleaseVersion>15.2.2</ReleaseVersion>
+		<ReleaseVersion>15.3.0</ReleaseVersion>
 	</PropertyGroup>
 	<ItemGroup>
 		<PackageReference Include="UDS.Net.Dto" Version="10.1.0" />
--- a/src/UDS.Net.Web.MVC.Services/UDS.Net.Web.MVC.Services.csproj
+++ b/src/UDS.Net.Web.MVC.Services/UDS.Net.Web.MVC.Services.csproj
@@ -4,13 +4,13 @@
 		<TargetFramework>netstandard2.1</TargetFramework>
 		<OutputType>Library</OutputType>
 		<PackageId>UDS.Net.MVC.Services</PackageId>
-		<Version>15.2.2</Version>
+		<Version>15.3.0</Version>
 		<Authors>Sanders-Brown Center on Aging</Authors>
 		<Description>Implemented service layer for using MVC front-end with API back-end</Description>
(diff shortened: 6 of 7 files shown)


mlan225:
Issue has been resolved, can go into detail on the resolution with the team. My previous post was inaccurate and has been hidden to avoid confusion.

### Visual aid for resolution discussion
[C2T section 13 bug.pdf](https://github.com/user-attachments/files/23257304/C2T.section.13.bug.pdf)

mlan225:
Looking into the C2 to verify its functionality and make slight changes to the view
- [ ] Verify the functionality presented in the instruction text, and understand how the form is conducted
  - Do the REY1REC sections need to function the same on both in-person and telephone visits? 
    -  Currently, the instruction text seems to provide different instruction for skip logic in the in-person and telephone. Look into this. 
  - Look back through older conversations around the C2 form to verify or discover functionality
- [ ] If possible, create a partial to reuse the REY1REC sections on in-person and telephone visits
- [ ] Make sure the variables in the REY1REC section are visible as it is in the in-person view within this PR

mlan225:
Tests failing on github action that were not failing on local, will need to look into this

ashleybot (changes requested):
@mlan225 This is looking awesome! Some big improvements here. Some clean-up items before I continue reviewing.

- [x] Rename _C2Layout.cshtml to _LayoutFormBase.cshtml and clean and reorganize code for reusability. This will be easier to maintain long-term. (see diagram below). _LayoutForm.cshtml should use _LayoutFormBase.cshtml (see example below).
- [x] Remove the turbo-frame in C2.cshtml. You need them in the partials so that Turbo can replace content correctly on post back, but having it in the page as well causes it to be duplicated and will cause issues (see screenshot of rendered html and diagram example below).

### Diagram of layouts, page, and partials
Image

### Rendered code with duplicated turbo-frame
Image

### Example of changes to _LayoutForm.cshtml
Screenshot 2025-12-03 at 12 21 07 PM

src/UDS.Net.Forms/Pages/UDS4/C2.cshtml.cs
```diff
@@ -637,7 +637,7 @@ public async Task<IActionResult> OnPostChangeMode()
 
             //Apply change to mode and modality based on form switch data
             C2.MODE = modeSwitch;
-            C2.RMMODE = modalitySwitch;
+            C2.RMMODE = C2.MODE == FormMode.InPerson ? null : modalitySwitch;
```
mlan225: Applied logic here to set RMMODE to null when the automatic form change is submitted with MODE = InPerson. This is to avoid hanging on to the current RMMODE value when MODE is changed from Remote to InPerson

mlan225:
> @mlan225 Things are looking much better. I returned to the original bug in the card and the issue still persists.
> 
> Steps to replicate:
> 
> 1. Fill out any fields in the C2 for remote telephone with a default value, but follow specific steps for sections 6 and 13.
> 2. Select VERBALTEST = 1
> 3. Section 6: Questions 6a-6n, enter 15 for the trials and 0 for the intrusions (see screenshot)
> 4. Section 13:
> 5. Question 13a REYDREC = 15
> 6. Question 13b REYDINT = 0
> 7. Question 13c REYDTI = 25
> 8. Question 13d REYTCOR = 15
> 9. Question 13e REYFPOS = 0
> 10. Successfully finalize the form and return to visit details
> 11. Click back into C2 and confirm Section 13 has persisted with the correct values and loads properly with the fields enabled
> 12. Now change Question 6a REY1REC to 95 (see screenshot)
> 13. Scroll down and see that some fields in Section 13 are _incorrectly_ disabled. These should only be controlled by REYDREC.
> 14. Finalize and save the form
> 15. The form is returned with errors because the Section 13 fields have been incorrectly disabled and therefore not posted. Additionally, the values have been lost because they are now posted back as null.
> 16. Styling is incorrect because the only class on the input is now `input-validation-error`. We had some custom styles in uds.net.forms.css that were incorrectly removed in [6f86d8f](https://github.com/UK-SBCoA/uniform-data-set-dotnet-web/commit/6f86d8fe3aaf995250d90e812529b711a97327a0) @Oddvocado
> 
> ### Section 6 Trials = 15 and intrusions = 0
> Image
> ### Section 6 REY1REC = 95
> Image
> ### Section 13 incorrectly disabled
> Image
> ### After postback, Section 13 failure
> Image
> ### CSS class on input after postback (@Oddvocado)
> Image
> Looping in @Oddvocado so that he adds back in the custom css in his PR.

## Resolution

We've discussed this bug and implemented a resolution. REY1REC should NOT control the enable/disable states of the questions in Section 13 and the connection was removed

|Before| After|
|-------|-------|
|Applying 95-98 to REY1REC results in disable of section 13 questions. When section 13 has values previously, they are cleared (aside from REYDREC) and upon save trigger a validation error for all the cleared inputs| Applying 95-98 to REY1REC no longer controls disable state of section 13 questions. As a result, when section 13 has values previously, only the server side validation is triggered because the previous values are NOT cleared (for all except REYDREC)|
|image| image |
|image| image|
| image| image |

PR #592, opened 2025-11-05
A1 Follow-up Visit 

### LINKED ISSUE #411: A1 Follow-up
Implement display of input fields for follow-up visits and adjust required variables.

References

- PDF => https://github.com/naccdata/uniform-data-set/blob/main/forms/uds/a1/UDS4-FVP-Form-A1-FinalProof.pdf
- Variable list => https://github.com/naccdata/uniform-data-set/blob/main/forms/uds/a1/form_a1_fvp_questions_and_vars.csv

CallowaySimon:
Resolves #411 

## A1 Follow-up Visit

Created A1 Follow-up Visit form. Adjusted Required Variables.

Added functionality to RequireOnFinalized Attribute so that Packet Kind can optionally be passed to only require for specific visits, or left blank to be required for all kinds.

- [x] Adjusted question numbering
- [x] Adjusted required variables
- [x] Adjusted instructions and demographics paragraph

### DIFF
--- a/src/UDS.Net.Forms.Tests.Runtime/UDS.Net.Forms.Tests.Runtime.csproj
+++ b/src/UDS.Net.Forms.Tests.Runtime/UDS.Net.Forms.Tests.Runtime.csproj
@@ -4,7 +4,7 @@
 		<TargetFramework>net8.0</TargetFramework>
 		<Nullable>enable</Nullable>
 		<ImplicitUsings>enable</ImplicitUsings>
-		<ReleaseVersion>15.1.2</ReleaseVersion>
+		<ReleaseVersion>15.1.3</ReleaseVersion>
 	</PropertyGroup>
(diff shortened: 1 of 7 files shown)


ashleybot (changes requested):
@CallowaySimon This is great! Improving the required on finalized tag helper to support specific packet types will set us up with a great foundation for these follow-up forms. Some tiny coding convention things:

- [x] Pascal case for public properties (camel case is used in parameter and local variable naming)
- [x] Prepend partial view names with an underscore. Helps to easily differentiate partials from full views.

### Reference
https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/coding-style/identifier-names#naming-conventions

https://learn.microsoft.com/en-us/aspnet/core/mvc/views/partial?view=aspnetcore-9.0#declare-partial-views

PR #599, opened 2025-11-07
Reload bug fixed

### LINKED ISSUE #593: A3 bug on load
**Describe the bug**
When navigating to A3 from visit details sometimes the js has errors and prevents fields from being correctly disabled.

Image

ahmedhaly7777:
Fixes #593 

**Stimulus implementation tests:**

- [ ] Biological parents section saving, validation, and disabling works as expected.
Screenshot 2025-11-13 092935
Screenshot 2025-11-13 093619

- [ ] SIBS and KIDS fields control the enabling of the rows
Screenshot 2025-11-13 093746

- [ ] SIBS and KIDS rows validation works as expected
Screenshot 2025-11-13 094049

**Potential bug:**

- [ ] Validation error messages still appear on disabled rows that were previously enabled. (This is also happening in the test env) But, if the user completes the enabled fields and finalizes, these errors do not prevent the form from being finalized.
Screenshot 2025-11-13 095146

### DIFF
--- a/src/UDS.Net.Forms/Pages/UDS4/A3.cshtml
+++ b/src/UDS.Net.Forms/Pages/UDS4/A3.cshtml
@@ -62,7 +62,7 @@
                                 <div class="grid grid-cols-2 sm:grid-cols-1">
                                     <div class="sm:col-span-1">
                                         <p class="mt-1 text-sm leading-6 text-gray-400">MOMYOB</p>
-                                        <input asp-for="@Model.A3.MOMYOB" placeholder="YYYY" />
+                                        <input asp-for="@Model.A3.MOMYOB" placeholder="YYYY" data-a3-enable-intra-row-target="input" />
                                     </div>
                                 </div>
                             </fieldset>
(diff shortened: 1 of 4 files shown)


src/UDS.Net.Forms/wwwroot/js/stimulus_controllers.js
```diff
@@ -16,6 +16,8 @@ import a4 from "./js_controllers/a4_controller.js"
 import modal from "./js_controllers/modal_controller.js"
 import checkboxButton from "./js_controllers/checkboxButton_controller.js"
 import pageRefreshOnSubmit from "./js_controllers/pageRefreshOnSubmit_controller.js"
+import EnableRows from "./js_controllers/EnableRows_controller.js"
+import EnableIntraRow from "./js_controllers/EnableIntraRow_controller.js"
```
ashleybot: @ahmedhaly7777 We need to code standards and naming conventions for each library. For Stimulus, filename words should be separated by underscores and lowercase. So `EnableIntraRow_controller.js `should be `enable_intra_row_controller.js`

And that means the identifier shoudl be `enable-intra-row` rather than `EnableIntraRow`.

Can you also make the naming A3 specific? Maybe `a3_rows.js` and `a3-rows`, and `a3_intra_row.js` and `a3-intra-row`?

ashleybot (changes requested):
@ahmedhaly7777 With Stimulus, we should not need vanilla js DOM selectors and event handlers.

- [ ] DOM selectors should be targets.
- [ ] Event handlers should instead be actions.

### Reference

- Stimulus naming conventions for filenames and identifiers https://stimulus.hotwired.dev/reference/controllers#modules
- Targets https://stimulus.hotwired.dev/reference/targets
- Actions https://stimulus.hotwired.dev/reference/actions

src/UDS.Net.Forms/Pages/UDS4/A3.cshtml
```diff
@@ -358,22 +357,26 @@
...
-    <div>
+    <div data-controller="a3-enable-rows" data-a3-enable-rows-max-rows-value="15">
         <h3 class="mt-16 text-base font-semibold leading-7 text-gray-900">Section 3. Biological children</h3>
 
         <div class="sm:grid sm:grid-cols-3 sm:items-start sm:gap-4 sm:py-6">
             <label asp-for="A3.KIDS"><span class="counter"></span> @Html.DisplayNameFor(m => m.A3.KIDS)</label>
             <div class="mt-4 sm:col-span-2 sm:mt-0">
                 <p class="text-sm text-gray-500" asp-description-for="@Model.A3.KIDS"></p>
-                <input asp-for="@Model.A3.KIDS" />
+                <input id="A3_KIDS" asp-for="@Model.A3.KIDS" data-a3-enable-rows-target="countInput" data-action="input->a3-enable-rows#updateRows change->a3-enable-rows#updateRows" />
```
ashleybot: @ahmedhaly7777 I think you should be able to use event shorthand here like so `a3-enable-intra-row#handlePrimaryChange`

https://stimulus.hotwired.dev/reference/actions#event-shorthand

Image

src/UDS.Net.Forms/wwwroot/js/js_controllers/a3_enable_intra_row_controller.js
```diff
@@ -0,0 +1,54 @@
...
+        });
+    }
+
+    updateFields(primaryDx) {
+        const value = primaryDx.value.trim();
+        const primaryValue = parseInt(value, 10);
+
+        const shouldDisable = primaryDx.disabled || value === '' || isNaN(primaryValue) || primaryValue === 0 || primaryValue === 99;
+
+        const row = primaryDx.closest('tr');
```
ashleybot: @ahmedhaly7777 You shouldn't need to use vanilla DOM selectors like closest. The controller is set on the tr, so primaryDx is the input for that tr. dependentTargets is the collection of targets for that tr.

ashleybot (changes requested):
@ahmedhaly7777 I think setTimeout is only needed because there are some bugs on the page.

This naming convention used to reference the outlet is incorrect:
`data-a3-enable-rows-a3-enable-intra-row-outlet=".a3-intra-row-outlet"`

It should be something more like `data-a3-enable-rows-outlet=".a3-intra-row-outlet"`

I'm noticing that the html is also unusual. The page has a tr and a single td and then the partial has a usual tr with the correct number of td. We shouldn't wrap a table row in a another table row. I'm surprised the formatting isn't more revealing in the browser.

```
                        <tr data-enable-rows-target="row">
                            <td>
                                @Html.EditorFor(m => m.A3.Children[i])
                            </td>
                        </tr>
```

Image

I added a commit to further show that setTimeout is only needed because of bugs. I commented out the setTimeout code and added some debug logs. You can see that the UI behavior still works as intended. There is a Stimulus warning that provides more details on bugs probably caused by naming convention errors:

Image

Next steps:
1. I'd like to see a diagram of how the html and stimulus js are working on this page. Sketch out all the flows that can happen on this page.
2. Add some more debug logging to double check your diagram against what is really happening in the code.
3. Cleanup the code (remove commented out code and logging) and fix bugs. Pay attention to detail to make sure the proper naming used.

Issue #603, opened 2025-11-14
Display more error and alert information on packet submission details


ashleybot:
Display more relevant information on packet submission details like old value and current value, and separate error count total into unresolved, resolved, and ignored.

Here are some examples:

<img width="1095" height="1106" alt="Image" src="https://github.com/user-attachments/assets/18ed54bd-2f0d-48cd-b838-3c58c2d634a2" />

**Bonus**
Look for more data that can be displayed like maybe when it was resolved or ignored. Also, we are running out of real estate here so look for some columns that could be combined. For example, perhaps Assigned To is only displayed if it has not been resolved or ignored yet? Maybe location can be displayed along with form?

mlan225:
In terms of creating screen space, we might be able to merge assigned to, resolved by, and ignored by information into 1 column called "status" 

### Status
- Assigned to: user@uky.edu --> unresolved
- Resolved by: user@uky.edu --> resolved
- IgnoredBy: user@uky.edu --> ignored

Applying location with the form sounds like a good idea, I had initally thought we could color code the link to the form (and maybe include a symbol) to the form link, however, this would be less intuitive than text and may confuse some users

PR #609, opened 2025-12-01
Display current form values in packet submission error table

mlan225:
Resolves: #603 

## Add information to the packet submission table
Add and replaced columns on the packet submission table in order to display more helpful information to the user

To help get the testing started faster, you can find a dummy error submission file Here: [error return test(in).csv](https://github.com/user-attachments/files/24065649/error.return.test.in.csv). 

> ⚠️ Important: The value in the file column will need to have the number appended to "NACC" and before "_FORMS-VISIT-1_UDS". An example number to test with locally can be obtained from our Teams fileshare

### Updated UI
Packet submission flow

### flow with categories 
Packet submission flow example

### UI Before and after
| Before | After |
|---------|------|
| Packet submission before | Packet submission after |

#### Changes from before to after
- Form column and location have been combined into form column
- Assigned to and resolved by have been combined into a status column
- Date Created column has been added
- Previous value column has been added
- Current value column has been added
- Categories with counts have been added to replace the total count
- Filter has been added through within the category count

### DIFF
--- a/src/UDS.Net.Forms.Tests.Runtime/UDS.Net.Forms.Tests.Runtime.csproj
+++ b/src/UDS.Net.Forms.Tests.Runtime/UDS.Net.Forms.Tests.Runtime.csproj
@@ -4,7 +4,7 @@
 		<TargetFramework>net10.0</TargetFramework>
 		<Nullable>enable</Nullable>
 		<ImplicitUsings>enable</ImplicitUsings>
-		<ReleaseVersion>15.3.6</ReleaseVersion>
+		<ReleaseVersion>15.3.7</ReleaseVersion>
 	</PropertyGroup>
 	<PropertyGroup Condition=" '$(RunConfiguration)' == 'https' " />
 	<PropertyGroup Condition=" '$(RunConfiguration)' == 'http' " />
--- a/src/UDS.Net.Forms.Tests/UDS.Net.Forms.Tests.csproj
+++ b/src/UDS.Net.Forms.Tests/UDS.Net.Forms.Tests.csproj
@@ -5,7 +5,7 @@
 		<ImplicitUsings>enable</ImplicitUsings>
 		<Nullable>enable</Nullable>
 		<IsPackable>false</IsPackable>
-		<ReleaseVersion>15.3.6</ReleaseVersion>
+		<ReleaseVersion>15.3.7</ReleaseVersion>
 	</PropertyGroup>
 	<ItemGroup>
 		<PackageReference Include="coverlet.collector" Version="6.0.4">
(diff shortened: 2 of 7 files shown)


src/UDS.Net.Forms/Models/PageModels/PacketPageModel.cs
```diff
@@ -38,7 +38,7 @@ public async Task<IActionResult> OnGetAsync(int? id)
             if (id == null || id == 0)
                 return NotFound();
 
-            var packet = await _packetService.GetById("", id.Value);
+            var packet = await _packetService.GetPacketWithForms("", id.Value);
```
mlan225: An update from GetById to GetPacketWithForms was made here and in other packet submission and packet submission error-related areas in order to provide data for the `Current Value` logic. Without it, the data for packets in the model does not contain the form input data

mlan225:
### Failing Test
The `CheckForForms` method seems to have failed in the PR testing (but not local). I have a code change to this test that seems to allow it to succeed in my C2 issue card on PR test runs. I switched the assert statements to expect statements in the testing method and used the `await Page.WaitForRequestFinishedAsync()` method instead of the `WaitForURLAsync()`

I wanted to avoid repeating code changes from another PR. So I will move this into review for now and if we want the test code copied over from there, then I can have that done.

Here is the test as it is currently in this PR without changes (VisitTests / CheckForForms)

image

src/UDS.Net.Forms/Models/PageModels/PacketPageModel.cs
ashleybot: I don't think we should get every single form's data when the errors may only exist in 1 or 2 forms. For example, in the sample error file you shared, there were only errors in A1a and B4. Getting all forms here instead of only the 2 is a lot of data.

src/UDS.Net.Forms/Pages/PacketSubmissions/_Index.cshtml
```diff
@@ -3,6 +3,10 @@
     var allErrors = Model.PacketSubmissions
         .SelectMany(s => s.Errors)
         .ToList();
+
+    int pendingCount = allErrors.Where(s => s.Status == PacketSubmissionErrorStatus.Pending).ToList().Count();
```
ashleybot: @mlan225 To get  a count of items, you should not need to convert it to a list and then count. You should be able to do `allErrors.Where(...).Count();`

ashleybot (approved):
@mlan225 Looks good! Good to bump to next stable patch, squash-n-merge to main, and tag the commit to main with version.

PR #625, opened 2025-12-23
B4 row border styling is confusing

### LINKED ISSUE #623: B4 row border styling is confusing
Notice in the B4 that the row borders are above the inputs. This causes the rows to be offset in an unusable way

Image

Oddvocado:
Resolves #623 

- [x] Started PR in draft.
- [x] Captured screenshots 💻 
- [x] Wrote a great PR description (see [here](https://github.com/UK-SBCoA/COA/wiki/PR-Standards#pr-description)) 🦖 

This PR fixes misaligned tables in Sections 1 and 2 of the B4 form. There are no functional changes; all updates are limited to the UI. The stimulus controller and validation behavior should remain unchanged.

To align the table rows, I replaced the previous approach of using the ```<table-cell-radio-buttons>``` tag helper to create radio buttons in a new ```<tr>``` tag. Instead, each possible value now has its own radio button wrapped in a ```<div>```, which is then placed within the appropriate table cell.

| Before | After |
| ------- | ------ |
| image | image |

### DIFF
--- a/src/UDS.Net.Forms.Tests.Runtime/UDS.Net.Forms.Tests.Runtime.csproj
+++ b/src/UDS.Net.Forms.Tests.Runtime/UDS.Net.Forms.Tests.Runtime.csproj
@@ -4,7 +4,7 @@
 		<TargetFramework>net10.0</TargetFramework>
 		<Nullable>enable</Nullable>
 		<ImplicitUsings>enable</ImplicitUsings>
-		<ReleaseVersion>15.3.3</ReleaseVersion>
+		<ReleaseVersion>15.3.4</ReleaseVersion>
 	</PropertyGroup>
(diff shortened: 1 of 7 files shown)


ashleybot (changes requested):
@Oddvocado The border issue outlined has been resolved, but there are some other things that should now be fixed.

- [ ] Revert the headings back to their previous styling. Sometimes we want something more attention grabbing, but in this case, we don't need it.
- [ ] The borders are fixed, but the radio inputs now have a variable vertical alignment. This reduces usability. All inputs in the same row should be on the same baseline.
- [ ] The center alignment also affects usability because it is not following consistency within our app of left alignment with the label. The inputs should be left aligned.

Image

PR #626, opened 2025-12-23
Administration Format

CallowaySimon:
- [x] Started PR in draft.
- [x]  UI changes? Extensively tested with browser dev tools: inspect the markup, look at the console logs, and network requests. 📈
- [x]  Ran unit or Playwright tests locally (if applicable). 🧪
- [x] Captured screenshots 💻
- [x]  Wrote a great PR description (see [here](https://github.com/UK-SBCoA/COA/wiki/PR-Standards)) 🦖
- [x]  Marked PR as ready for review and moved into 👀 In review

## Adds the ADMIN fields to A1 and A1a

Added Administration Format to A1 and A1a initial and follow-up forms footer.

image

### Review Checklist
- [x] Administration field has options self and staff on A1 and A1a forms
- [x] Field is required
- [x] Field is disabled and not required on other forms
- Tested saving and finalizing forms
- Tested packet export

### DIFF
--- a/src/UDS.Net.Forms.Tests.Runtime/UDS.Net.Forms.Tests.Runtime.csproj
+++ b/src/UDS.Net.Forms.Tests.Runtime/UDS.Net.Forms.Tests.Runtime.csproj
@@ -4,7 +4,7 @@
 		<TargetFramework>net10.0</TargetFramework>
 		<Nullable>enable</Nullable>
 		<ImplicitUsings>enable</ImplicitUsings>
-		<ReleaseVersion>15.3.6-preview.1</ReleaseVersion>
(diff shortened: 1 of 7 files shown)


ashleybot (changes requested):
@CallowaySimon We should add a "--" item so that when a new A1 or A1a is created, the user is forced to select either Self or Staff. Having a default value here would probably cause some user entry error.

Image

Along these same lines, for A1a only, if the form mode is `Not Completed`. Should admin be exported as null or 2? If so, we might should auto-select administration if form mode is Not Completed.

CallowaySimon:
image

- Added -- option

image

- A1a autoselects value 2 when NotCompleted

PR #674, opened 2026-03-26
On packet finalization, recheck RxNorm CUIs

### LINKED ISSUE #605: On packet finalization, recheck RxNorm CUIs
We are seeing 3mo+ old packets that have RxCUIs that are no longer current. NACC does not allow these packets to be submitted.

Add a check on packet finalization for all RxCUIs. If any are no longer current display as a packet error and do not allow finalization.

Image

Note: This may require and update of the underlying RxNorm API client wrapper.

Oddvocado:
Resolves #605 

This PR relies on [this RxNorm PR](https://github.com/UK-SBCoA/rxNorm_dotnet/pull/19)

- [x] Started PR in draft.
- [x] Captured screenshots 💻 
- [ ] Wrote a great PR description (see [here](https://github.com/UK-SBCoA/COA/wiki/PR-Standards#pr-description)) 🦖 

This PR validates RXNorm CUI's from the A4 form on packet finalization. If a RxNorm CUI is invalid at the time of packet finalization, a warning is displayed to the user

Example:
image


ashleybot (changes requested):
@Oddvocado This should not search in the local cache for valid RxNormIds, but use the RxNorm API client library to search the currently active RxNorm drugs.

Here's the client library => https://github.com/UK-SBCoA/rxNorm_dotnet

It may need to be updated to use this endpoint. https://lhncbc.nlm.nih.gov/RxNav/APIs/api-RxNorm.findRxcuiById.html You may need to implement a new method that uses this endpoint (pull down the source code for that repo).

That endpoint will return Concept status when you give it an RxCUI/RxNormId. Concept statuses:
* Active
* Obsolete
* Remapped
* Quantified
* NotCurrent
* Unknown

If you get back a status for anything except "Active" then an alert should be displayed on packet finalization.

Oddvocado:
I've updated the RxNorm API to retrieve the value an RxNorm status. The UDS app has also been updated to use this method and validate RxNormId's. Ready for another review!

src/UDS.Net.Forms.Tests.Runtime/Services/LookupService.cs
```diff
@@ -10,7 +9,15 @@ public class LookupService : ILookupService
         public LookupService()
         {
         }
+        public Task<bool?> RxNormIsActive(string rxCUI)
+        {
+            throw new NotImplementedException();
```
ashleybot: @Oddvocado Instead of throwing not implemented, I wonder if it would be better to always return a true or "Active" status so that it would not prevent any tests that are testing packet finalization. #678

src/UDS.Net.Services/DomainModels/Visit.cs
```diff
@@ -483,6 +484,33 @@ public IEnumerable<VisitValidationResult> GetModelAlerts()
                 }
             }
 
+            if (a4g != null)
+            {
+                A4GFormFields a4Ds = a4g as A4GFormFields;
```
ashleybot: @Oddvocado Let's keep the a4g named as "a4" rather than changing the meaning with "a4Ds" and causing a repeat later `a4Ds?.A4Ds`

ashleybot (changes requested):
@Oddvocado A tiny variable name tweak and go ahead and create a preview and feature branch in our local repo since this will require the service methods to be updated there as well. Make it a minor version bump to `16.3.0-preview.*`

src/UDS.Net.Services/DomainModels/Visit.cs
Oddvocado: Updated!

src/UDS.Net.Forms.Tests.Runtime/Services/LookupService.cs
Oddvocado: RxNormIsActive has been updated to return "true" and GetRxNormStatus now returns "Active". Both methods have been made async as well

Oddvocado:
Requested changes have been made and a [PR has been created in the local UDS app](https://github.com/UK-SBCoA/COA.UDSv4/pull/232).

ashleybot (approved):
@Oddvocado This looks good! This will be a major version bump since we're creating a breaking change in the service layer. Update our COA.UDSv4 with the preview pkgs and get that PR approved before merging this to main.

Successfully alerted me to a non-active drug:

Image

ashleybot (approved):
@Oddvocado Good to move to 16.3.0 and merge to main!

Oddvocado:
@ashleybot  after our discussion about nullable return types. I removed the ability for RxNorm API service methods "RxNormIsActive" and "GetRxNormStatus" to return null. Both methods will now throw an argument exception and an error message informing the user the rxCUI parameter cannot be null or whitespace. Both the UDS web and our local UDS app have had their RxNorm dependency updated to v1.0.8-preview.2

src/UDS.Net.Web.MVC.Services/LookupService.cs
```diff
@@ -184,6 +183,25 @@ public async Task<List<RxNorm>> LookupRxNormApproximateMatches(string searchTerm
             return new List<RxNorm>();
         }
 
+        public async Task<bool> RxNormIsActive(string rxCUI)
+        {
+            if (!String.IsNullOrWhiteSpace(rxCUI))
+            {
+                var isActive = await _rxNormClient.RxNormIsActiveAsync(rxCUI);
```
ashleybot: @Oddvocado Does this response need to be cast? Could you just do `return await _rxNormClient.RxNormIsActiveAsync(rxCUI);` ?

ashleybot (changes requested):

src/UDS.Net.Web.MVC.Services/LookupService.cs
Oddvocado: Yeah for sure! I've gone ahead and removed the variable assignment.

ashleybot (approved):
@Oddvocado I added a commit to clean-up the code. It looks like it should've been cleaned up when the separate follow-up objects were reverted. Anytime reflection is being used when it doesn't need to be we should avoid it.

I also fixed the variable naming. I'm not sure what happened there.

Image

I added a test non-active drug in the reference to make it easier to test the alert scenario.

Image

Image

Run through all your tests again and if they look good then this is good to make stable and merge to main.

PR #684, opened 2026-04-13
Adjust export value application (current value, code, or null) to A3. Null out A3 sections when changes are not detected in section (parent, sibling, kid)

mlan225:
Resolves: #679 

### Export FVP variables need to be nulled out if no changes detected
The existing export was still using previous value codes even when no changes detected. Josh was able to confirm that values in sections need to be null if no changes detected. 

This also applies to kids and sibling count questions, as sections are to be skipped if no changes detected from previos visit 

| Section | Screenshot|
|--------|-------------|
| Parents section | image|
| Siblings section | image|
| Kids section | image|

### Null values when sections unchanged VS changed

| Change status | Screenshot |
|---------------|---------------|
| No sections changed, sections are null including sibs and kids counter when related sections are unchanged | image|
| Sections were changed, change codes are applied and sibs and kids counter inputs are written | image|


mlan225:
While confirming NACC validation, Josh was able to confirm that the export logic in this packet was necessary for a successful submission

src/UDS.Net.Forms/Pages/PacketSubmissions/Export.cshtml.cs
```diff
@@ -443,7 +441,8 @@ private async Task WritePacketDataAsync(CsvWriter csv, PacketSubmission packetSu
                         siblings[siblingsIndex].ETSEC = CompareA3Values(previousA3Fields.SiblingFormFields[siblingsIndex].ETSEC, siblings[siblingsIndex].ETSEC, "66", A3Section.Sibling);
                         siblings[siblingsIndex].MEVAL = CompareA3Values(previousA3Fields.SiblingFormFields[siblingsIndex].MEVAL, siblings[siblingsIndex].MEVAL, 6, A3Section.Sibling);
                         siblings[siblingsIndex].AGO = CompareA3Values(previousA3Fields.SiblingFormFields[siblingsIndex].AGO, siblings[siblingsIndex].AGO, 666, A3Section.Sibling);
-                    };
+                    }
```
ashleybot: @mlan225 C# coding conventions for syntax should keep semicolons on the same line as closing brackets.

src/UDS.Net.Forms/Pages/PacketSubmissions/Export.cshtml.cs
```diff
@@ -473,7 +505,7 @@ private async Task WritePacketDataAsync(CsvWriter csv, PacketSubmission packetSu
                     {
                         if (prop.Name != "FamilyMemberIndex")
                         {
-                            csv.WriteField(prop.GetValue(sibling));
+                            csv.WriteField(currentA3Fields?.NWINFSIB == 0 ? null : prop.GetValue(sibling));
```
ashleybot: @mlan225 This comparison could get tricky if we need to do yet another thing. CompareFollowUpValues() should be improved to support an additional optional parameter to control if the export value gets set to `null`. This will make it reusable for all follow-up forms.

ashleybot (changes requested):
@mlan225 Let's use this opportunity to clean up the code as well. Let me know any questions!

Image

mlan225:
@ashleybot and I spoke about an update to the A3 export functionality that will be taken into consideration during PR changes

### We will need to know if there changes in the entire section before we can know if we need to export a code or NULL
> Thes thoughts apply to all sections with previous visit codes (parents, siblings, kids) but uses siblings as the example.

| section | description | screenshot |
|---------|-------------|-------------|
| Iterate through siblings to decide the export value | If there were no changes in the first iteration but changes in the last iteration of the loop, the section will not be marked as changed until the last iteration. The first iteration will have no knowledge of this on the first loop through siblings and won't know to use NULL instead of error codes for the first iteration| image |
| Iterate through siblings to write data | Once we have iterated through all the sibling properties, we will know if there were any changes. The issue with only writing here is that we don't know about changes until we've done the initial loop through the section. We may have to conditionally write NULL here as we write data. | image |

mlan225:
### Changes 
- [ ] Added functionality for sections to be NULL when changes were not made, only adding previous value codes for changed sections
- [ ] Added Sibling and Kids counts to logic for previous value codes. Will be NULL as well when section is unchanged
- [ ] Change tracking properties were removed from class member scope and are now handled within the write method using currentA3FormField properties instead of variables
- [ ] Both compareA3Values methods were removed in place of the existing (and renamed) CompareFollowUpValue (renamed to GetExportObject)

## Review discussion questions
1. While class property variables were removed to combat the code smell, I feel as if there is still a lot happening here in order to make sure every property is handled properly. There are a lot of variables being created within the write method to handle comparing values and detecting changes
2. The compareA3Values methods were removed in order to use the single existing method. The existing method was adjusted to return an object with value and hasChanged properties to cut down on code for setting changed properties (specifically in the A3). I feel like this method is kind of doing 2 jobs with a focus on one form (for now), however, this seemed like an effective way to what we needed while leveraging a singular method.
3. As discussed previously, we can address the change properties ONLY AFTER we have looped through all properties to detect if a change has occured within each section. Due to this, the change property is still involved in a conditional when writing the value to the CSV after all properties have been looped.

### Feature workflow

a3 export flow

### CompareFollowUpValue changed to GetExportObject

| screenshot | description |
|------------|-------------|
| image| This method now returns an object with value and hasChanged. The value will be used for returning the property value after comparison and the hasChanged will be used to detect changes for change prop |

### Sibling and kids count handled during parent value checks

| screenshot | description |
|------------|-------------|
| image| Due to sibling count and kids count existing in the A3FormFields class, it is being handled after the parent props in the parent section |

### Manually looping through currentA3Fields class instead of running CSV.WriteRecord on it

| screenshot | description |
|------------|-------------|
| image| The CSV writer loops through the primitive properties of a class when running writeRecord() on an object. For more control over setting the properties to NULL after checking the change property I am manually implementing this loop. Similar to how sibs and kids are looped. |

mlan225:
### CSV compare results

Same count of rows with varying null value cells, which is to be expected since I used my local uds export against the uds4 test environment export

image

mlan225:
Josh was able to find out what to adjust on our export files to satisfy most of the requirements for the sandbox.

Some fields are difficult to satisfy and avoid errors / alerts due to it expecting "The same answer as previous visit", but we are unable to know what value it is expecting since we can't see the previous visits in the sandbox database at the moment.

Made an adjustment based on the export results:

- [x] Fixed bug with ETPR and ETSEC export values, it was exporting a single digit instead of the leading 0 string

src/UDS.Net.Forms/Pages/PacketSubmissions/Export.cshtml.cs
mlan225: This was discussed and a detailed comment was made below. In short, we need to loop through properties first before we can confirm that changes were present

mlan225:
Josh was able to get a successful sandbox submission with the export file from this feature ✅

mlan225:
Spoke with @ashleybot about using methods on the A3FormFields / SiblingFormFields / KidsFormFields classes to handle returning objects for export to use with CSV.Helper's WriteRecord() method. 

This would remove the responsibility for setting encoded values and checking against change properties in the export file over to the respective formField classes.

mlan225:
### Discussion topics

- [x] 	Go over flow chart
- [x] 	Talk about the decision to have A3FormFields call A3FamilyFormFields
  - They form fields and family form fields are associated with each other
  - No need for both class methods to be called from the central export file (that would be more work in the export)
- [x] 	Talk about the exportValueHelper
  - Maybe it could use a name change?
  - Supplies 2 methods to hand ints and strings of the formfield properties
- [x] 	Talk about the changeProperty delegate used in the exportvalueHelper methods
  - In order to keep track of changes without adding code fluff, the method can use a delegate to adjust section change properties WHILE it gets the export value
- [x] 	Removed early dev artifacts
  - Deleted the exportObject class that had changed and value props when getting export values
  - Set the old CompareFollowUpValues method back to how it was, since its being used in other forms in the export

### A3 export flow
A3_export_flow drawio

ashleybot (changes requested):
@mlan225 How are initial visit exports handled?

mlan225:
> @mlan225 How are initial visit exports handled?

It seems i've overwritten this catch in some recent changes. However, I've provided the adjustment the way it should have been. The export values are now applied if a previous visit can be found = follow-up visit. 

This change addresses the fact that writing data was only done when getting export values. Now writing occurs regardless and export values are applied when applicable.

### Adjustment flow

A3_export_initial_followup_adjustment drawio

### Self review of adjusted code
- [x] Initial export tested to export A3 values correctly
- [x] Follow-up export tested to export A3 values correctly

src/UDS.Net.Forms/Pages/PacketSubmissions/Export.cshtml.cs
```diff
@@ -398,12 +386,9 @@ private async Task WritePacketDataAsync(CsvWriter csv, PacketSubmission packetSu
 
                 A3FormFields? currentA3Fields = a3.Fields as A3FormFields;
 
-                //Create lists for siblings and kids by casting a3.fields data (IFormFields to A3FormFields)
-                List<A3FamilyMemberFormFields> siblings = ((A3FormFields)a3.Fields).SiblingFormFields;
-                List<A3FamilyMemberFormFields> kids = ((A3FormFields)a3.Fields).KidsFormFields;
```
mlan225, line 403: Sibling form fields and kids formfields are included within `currentA3Fields`. They were no longer used in current code and removed as the code treats `currentA3Fields` as the data source during the write process

mlan225:
### Personal review
- [x] A3 export works correctly in a follow-up when sections are unchanged
- [x] A3 export works correctly in a follow-up when all sections are changed
- [x] A3 export works correctly in a follow-up when 1 section is changed
- [x] Values in the A3 export follow-up are correctly written to the export file
- [x] A3 export correctly sets change properties to NULL for initial
- [x] Walk through code and make sure variable names are explanatory, conditional logic is handled effectively, and logic is efficient
- [x] Run local tests

### Review changes
- Adjust typo in code comment
- Adjust export helper delegate argument to `newInformationPropSetter`
  - This delegate will call the set method for the new information property it is provided

ashleybot (approved):
@mlan225 Reviewed the code and everything looks good! This is a huge design improvement: each form object handles it's own jobs, you created a utility that can be reused by multiple forms, and the code is much more readable and maintainable.

Good to bump to next stable patch and merge to main. No need to update our local repo b/c we're waiting on A4a before we can export follow-up visits.

PR #686, opened 2026-04-14
A1a FVP Updates

### LINKED ISSUE #412: A1a Follow-up
Reference

- PDF => https://github.com/naccdata/uniform-data-set/blob/main/forms/uds/a1a/UDS4-FVP-Form-A1a-FinalProof.pdf
- Variables list => https://github.com/naccdata/uniform-data-set/blob/main/forms/uds/a1a/form_a1a_fvp_questions_and_vars.csv

ahmedhaly7777:
Fixes #412 

**Description:**

- [ ] Added new instructional paragraphs 
- [ ] Updated to new FVP paragraphs 
- [ ] Updated to new FVP questions 
- [ ] Updated to new FVP answers

**Note:**
Administration options are similar to the one in FVP:
admin

Instructions:
Page top instructions

Section 1:
section1 paragraph

Section 2:
section 2 paragraph

Section 3:
Section 3 paragraph

Section 4:
Section 4 paragraph

Section 5:
Section 5 paragraph

Question 2:
Q2 paragraph

Question 6:
Q6

Question 11:
Q11

Question 13:
Q13

Question 16:
Q16 paragraph

Question 20:
Q20

Question 25:
Q25

Question 36:
Q36

Question 37:
Q37

Question 38:
Q38

- [x] Started PR in draft.
- [ ] Ran unit or Playwright tests locally (if applicable). 🧪 
- [x] Captured screenshots 💻 
- [x] Wrote a great PR description (see [here](https://github.com/UK-SBCoA/COA/wiki/PR-Standards#pr-description)) 🦖 


src/UDS.Net.Forms/Pages/UDS4/A1a.cshtml
```diff
@@ -6,13 +6,33 @@
 }
 
 <div>
-    <p class="mt-1 text-sm text-gray-500">
-        <b>INSTRUCTIONS:</b> The following questions are designed to assess your current and past life experiences. These questions will help us understand how certain experiences affect your health.
-        You do not have to answer any question that makes you feel uncomfortable.
-    </p>
+    @if (Model.Visit.PACKET == PacketKind.F)
```
ashleybot: @ahmedhaly7777 Instructions should be the same for initial and follow-up packets.

Image

ashleybot (changes requested):
@ahmedhaly7777 Be sure to always reference the source of truth - NACC documentation.

A lot of these changes are not current and are only applied to follow-up packets when they should be applied to all packet types.

- [ ] Reference both PDFs and update code

### Reference

A1a initial PDF => https://files.alz.washington.edu/documentation/wm-uds4-ivp-a1a.pdf
A1a follow-up PDF => https://files.alz.washington.edu/documentation/wm-uds4-fvp-a1a.pdf

ahmedhaly7777:
- [ ] A1a IVP & FVP forms have been updated to be current with NACC's documentation.

- [ ] "Optional" label is added to A1a FVP.

Screenshot 2026-05-07 103314

Finalized

src/UDS.Net.Forms/Pages/UDS4/A1a.cshtml
ahmedhaly7777: Instruction paragraph is updated.

ahmedhaly7777:
@ashleybot Branch updated with main.

ashleybot (changes requested):
@ahmedhaly7777 Move more stuff to either A1a in Forms or Domain layer rather than modifying the _FormFooter.

ahmedhaly7777:
> @ahmedhaly7777 Move more stuff to either A1a in Forms or Domain layer rather than modifying the _FormFooter.

Moved `NotComplete` reason `Option ` to the domain for the A1a form.

ahmedhaly7777:
@ashleybot Ready for review

ahmedhaly7777:
@ashleybot  Ready for review!

src/UDS.Net.Services/DomainModels/Form.cs
```diff
@@ -91,7 +91,10 @@ private void SetFieldsBasedOnKind(PacketKind packetKind)
             }
             else if (Kind == "A1a")
             {
-                Fields = new A1aFormFields();
+                Fields = new A1aFormFields
```
ashleybot: @ahmedhaly7777 Instead of introducing a boolean, let's just pass packetKind into the constructor of A1aFormFields. This gives us more flexibility if they introduce more rules for different packet kinds, and we still get to have the solution at the same time.

It also makes the code more readable and safe to use because in the constructor, now we know that we must know which packet kind we are dealing with before we instantiate an A1a.

ashleybot (changes requested):
@ahmedhaly7777 Let's swap the boolean property for PacketKind enum. And keep an eye on version numbers when you merge in main, sometimes you need to bump your patch up above whatever was updated in main (I can see it was 18.0.5), so you need to go to a patch above that depending on what is available.

This will require some more changes in other files since we're modifying the constructor, but will be beneficial long-term.

Also check to make sure the Playwright Test for FormFooterNotCompletedTests testing these new scenarios added in this PR.

src/UDS.Net.Services/DomainModels/Form.cs
ahmedhaly7777: Playwright test added.

Remove the Boolean and pass `packetKind `into the constructor.

src/UDS.Net.Services/DomainModels/Forms/A1aFormFields.cs
```diff
@@ -92,6 +100,13 @@ public IEnumerable<AdministrationFormat> AdministrationFormats
             }
         }
 
+        public PacketKind PacketKind { get; }
+
+        public A1aFormFields(PacketKind packetKind)
```
ashleybot, line 105: @ahmedhaly7777 Let's always keep constructors grouped together (needs to be moved down). And we now have this new scenario where PacketKind is necessarily required by the A1a form, so the other constructors need to be updated as well. And we may need to remove the empty constructor to ensure the A1a always gets the information it needs about PacketKind.

ashleybot (changes requested):
@ahmedhaly7777 Try to look at this from a higher level, how should the A1a be used by other developers going forward?

There are some other tests that need to be addressed. I don't see a test that checks to make sure the new option is disabled for initial packets and enabled for follow-up packets.

And, for example, in the VisitFinalizable tests A1a forms should always receive PacketKind now.

Image

Keep an eye on version numbers. These version numbers are out of sync with main.

PR #691, opened 2026-04-20
A1 validation and priocc dropdown

### LINKED ISSUE #330: A1 Validation and provide autocomplete drop-down for PRIOCC
Provide autocomplete drop-down for PRIOCC so as the user types in an occupation it displays matching results. See RxNorm for example of UI interactions. Also validation on finalize against stored occupations.

CallowaySimon:
Resolves #330 
- [x] Started PR in draft.
 - [x] Captured screenshots 💻
 - [ ] Marked PR as ready for review and moved into 👀 In review
## A1 PRIOCC Autocomplete
Adds autocomplete to A1 PRIOCC field. Occupations are store as JSON with Code and Occupation name. Service methods for Autocompletion are added to Lookup Service.

image
Added a search field to PRIOCC. Either a code or a occupation name can be entered. An Occupation is then selected from the search. 

image
 The Readonly PRIOCC field is then set to the code for that occupation. The field can be cleared using the Clear button.


ashleybot (changes requested):
@CallowaySimon This is looking pretty good! Can we level it up to have only one UI field? You could still keep the PRIOCC field as a hidden input if needed.

Nice support of sub-string search

Image

### A4 RxNorm Lookup UI

1. User can begin typing and the list is displayed. Then the user can use keyboard or mouse to select something from the list.
2. Hitting enter or clicking on the item will autocomplete the text input.

| Step 1 | Step 2 |
| --- | --- |
| Image | Image |

### Occupations Lookup UI
There are some bugs with UI interactions and using the keyboard.

User begins typing

Image

Using keyboard down arrow and enter autocompletes with HTML rather than the occupation.

Image

Only entering the occupation code and hitting enter throws an error. This interaction should be allowed for super-users. (This is broken in RxNorm as well, but we can fix that separately).

Image

Image

CallowaySimon:
Updated Dropdown to use a single field.
image

Fixed input when selecting by pressing Enter
image
image

Also supported super user entering only occupation code

ashleybot (changes requested):
@CallowaySimon Can the existing autocomplete be upgraded rather than making one specific to occupations?

CallowaySimon:
AutoComplete now stays a reusable controller, and Rx norm and Occupation extend it for specific use cases while relying on the base controller where possible.

ashleybot (changes requested):
@CallowaySimon This is so nice! There is a little missing validation for PRIOCC. I was able to finalize an A1 with a made up "235325235" value. It was not persisted (the form reopened with the input blank), but as a user I wouldn't have know there was an issue.

Form finalizes with no validation errors with:
Image

Opens with:
Image


### UX is smooth!
I tested several different ways to choose PRIOCC, by the code or by the text value. And I also tested several select methods for both PRIOCC and A4 RxNorm lookup, arrow keys, enter and clicking - _everything_ has such a smooth natural UX!

CallowaySimon:
Added Validation for PRIOCC for Invalid Codes.

ashleybot (approved):
@CallowaySimon Good to bump to next stable patch and merge to main!

PR #692, opened 2026-04-23
Fix B3 and B6 when MODE=NotCompleted

### LINKED ISSUE #689: Can B3 be MODE=NotCompleted without including data?
Currently, it is not possible to choose MODE=NotCompleted and allow null fields in the remainder of the form. Read NACC official documentation and determine if this aligns with requirements for data entry and export. If this is by design, nothing needs to be done, if it is a bug then implement a fix.

### LINKED ISSUE #690: Can B6 be MODE=NotCompleted without including data?
Currently, MODE=NotCompleted requires a value for GDS. Review the official documentation and determine if GDS can be null if MODE=NotCompleted. If it should be set to 88, update the javascript to automatically set GDS to 88 if MODE=NotCompleted and handle if MODE is changed to InPerson or Remote afterward. Update Playwright Tests.

CallowaySimon:
Resolves #689
Resolves #690 
- [x] Started PR in draft.
 - [x] Captured screenshots 💻
 - [ ] Marked PR as ready for review and moved into 👀 In review
## Fixes Not Completed Issue with B3 and B6
Fixes issue with B3 and B6 where form mode could not be saved as NotCompleted without filling out all fields.
image
Both forms can now be saved as NotCompleted will all fields null. 
image


ashleybot (changes requested):
@CallowaySimon Instead of making a change in the RequiredIfAttribute (it will have a wide breadth of effect), I'd like to understand more about the pieces at play.

Can we first start with some Playwright tests that test the form footers in forms that can be  MODE=NotCompleted? They should test every NotCompleted applicable form for correct successful tests and correct failures.

And then after the tests are in place we can look at the wider issue. Diagrams will help.

Another thing that I'd like to tackle with this is refactoring the dropdownReset js. We need to make it specific for the form footer (it really isn't possible to use it anywhere else anyway).

CallowaySimon:
I added playwright tests for all forms. On non required forms, it tests whether the form can be saved as not completed without entering any data. It also tests validation on the Not Completed Reason field. On the Required forms it ensures that the NotCompleted Mode option is disabled.

I Added back the NotCompleted Check on the RequiredIf attribute. 

I also updated the A2 NEWINF validation to use the updated packet kind specific RequireOnFinalized attribute.

The A4a Allowed Modalities was updated, as it was allowing for NotCompleted when A4a is a required Form.

I also Refactored the dropdownReset controller to be specific to the Form Footer, renaming it formFooter_controller.js

ashleybot (changes requested):
@CallowaySimon Fantastic improvements! Should the C2 be included in the NotCompleted disabled tests? The Mode field is not in the form footer, but there is still a select box at the beginning of the C2 page.

Image

C2 is required => https://files.alz.washington.edu/documentation/udsv4-form-submission-guidance.pdf

CallowaySimon:
I have added a playwright test to ensure that NotCompeted Mode is disabled on the C2 form

ashleybot (approved):
@CallowaySimon Looks great! Bump to next stable patch and you can make a hotfix commit to update dependencies in our local UDS4.

PR #702, opened 2026-06-03
Adjust GetEncodedValue helper method to return code for null matches

### LINKED ISSUE #701: A3 provide encoded value on null match
The A3 form is requiring that the disabled fields of each section are encoded as previous value codes when previous visit value is NULL

mlan225:
Resolves: #701 

## Change the way that A3 handles NULL values between initial and follow-up visits
Instead of exporting value as NULL when the previous visit is NULL for ETSEC, MEVAL, and AGEO for mom, dad, sibling, and kid, we will provide a previous visit code when the section has changed

- [ ] Disabled values in changed sections export as encoded values instead of NULL
- [ ] Unprovided Kids and Sibs in changed sections remain NULL instead of using encoded values
- [ ] Unchanged sections continue to export NULL for sections that have not changed

#### Form view
image

#### Export view
image

> Instead of exporting NULL for MOMETSEC, MOMMEVAL, MOMAGEO previous value codes are provided

#### Form view
image

#### Export view (disabled values in siblings)
image

> If section is changed, then provide encode values for disabled fields when sibling was provided

#### Export view (unprovided siblings)
image

> Siblings that were not provided remain as NULL

### DIFF
--- a/src/UDS.Net.Services/DomainModels/Forms/A3FamilyMemberFormFields.cs
+++ b/src/UDS.Net.Services/DomainModels/Forms/A3FamilyMemberFormFields.cs
@@ -18,6 +18,7 @@ namespace UDS.Net.Services.DomainModels.Forms
             {
                 A3FamilyMemberFormFields encodedFamilyMemberFormFields = new A3FamilyMemberFormFields();
 
+                encodedFamilyMemberFormFields.FamilyMemberIndex = previousFamilyMemberFormFields.FamilyMemberIndex;
                 encodedFamilyMemberFormFields.YOB = ExportHelper.GetEncodedValue(previousFamilyMemberFormFields.YOB, this.YOB, 6666, hasNewInformation);
                 encodedFamilyMemberFormFields.AGD = ExportHelper.GetEncodedValue(previousFamilyMemberFormFields.AGD, this.AGD, 666, hasNewInformation);
                 encodedFamilyMemberFormFields.ETPR = ExportHelper.GetEncodedValue(previousFamilyMemberFormFields.ETPR, this.ETPR, "66", hasNewInformation);
--- a/src/UDS.Net.Services/DomainModels/Forms/A3FormFields.cs
+++ b/src/UDS.Net.Services/DomainModels/Forms/A3FormFields.cs
@@ -105,11 +105,33 @@ namespace UDS.Net.Services.DomainModels.Forms
                 //Encode siblings and kids
                 if (encodedFormFields.SiblingFormFields != null)
                 {
-                    encodedFormFields.SiblingFormFields = encodedFormFields.SiblingFormFields.Select((siblingFields, index) => siblingFields.GetEncodedFormFields(previousA3Fields.SiblingFormFields[index], hasNewInformation => encodedFormFields.NWINFSIB = hasNewInformation)).ToList();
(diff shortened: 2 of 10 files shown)


mlan225:
While applying the changes for this PR, I realized that changing the null match logic from "return NULL" to "return encoded value" will causes all the unprovided siblings and children to also return encoded values and NOT just the disabled specific fields in each row

Josh was able to test a packet for the 3 disabled values of each section following this logic and it passed validation. However, we did not check for supplying encoded values for each unincluded sibling and child question. 

We will need to validate this behavior OR implement further logic to detect when a child / sibling row is part of the count before returning encoded values on a NULL value match.

mlan225:
## Self review 📝 
- [x] Started PR in draft.
- [x] Captured screenshots 💻
- [x] Marked PR as ready for review and moved into 👀 In review

src/UDS.Net.Services/DomainModels/Forms/A3FamilyMemberFormFields.cs
```diff
@@ -18,6 +18,7 @@ public A3FamilyMemberFormFields GetEncodedFormFields(A3FamilyMemberFormFields pr
             {
                 A3FamilyMemberFormFields encodedFamilyMemberFormFields = new A3FamilyMemberFormFields();
 
+                encodedFamilyMemberFormFields.FamilyMemberIndex = previousFamilyMemberFormFields.FamilyMemberIndex;
```
mlan225, line 21: Added family member index back into the object on return. Not currently used, but added for the sake of returning all object data and if it would wish to be accessed in the future

src/UDS.Net.Services/DomainModels/Forms/A3FormFields.cs
```diff
@@ -105,11 +105,29 @@ private A3FormFields GetEncodedFormFields(A3FormFields previousA3Fields)
                 //Encode siblings and kids
                 if (encodedFormFields.SiblingFormFields != null)
                 {
-                    encodedFormFields.SiblingFormFields = encodedFormFields.SiblingFormFields.Select((siblingFields, index) => siblingFields.GetEncodedFormFields(previousA3Fields.SiblingFormFields[index], hasNewInformation => encodedFormFields.NWINFSIB = hasNewInformation)).ToList();
+                    encodedFormFields.SiblingFormFields = encodedFormFields.SiblingFormFields.Select((siblingFields, index) =>
+                    {
+                        if (index < (encodedFormFields.SIBS != 66 ? encodedFormFields.SIBS : previousA3Fields.SIBS))
```
mlan225, line 110: For kids and sibs, if the current encoded count property is 66 (provided at previous visit) then fall back to using the previous count.

src/UDS.Net.Services/Utilities/ExportHelper.cs
```diff
@@ -6,11 +6,6 @@ public static class ExportHelper
     {
         public static int? GetEncodedValue(int? previousValue, int? currentValue, int code, Action<int?> newInformationPropSetter)
         {
-            if (previousValue == null && currentValue == null)
-            {
-                return null;
-            }
-
```
mlan225, line 13: NULL matches returning NULL has been removed

Now NULL matches return an encoded value and the change property determines if the value will be an encoded value or NULL in the export

Oddvocado:
@mlan225 @ashleybot This looks good to me! I tested exporting an A3 form that matches the previous visit, all values were exported as NULL. I then tested an follow up A3 form with changed values and as expected, only the changed fields had their values exported and the unchanged fields were exported with the encoded value.

Oddvocado (approved):

PR #718, opened 2026-06-29
re-apply the null export when previous and current value are null

### LINKED ISSUE #717: Revert the A3 null export adjustments
We will be reverting the recent adjustments to the A3 export logic

when previous and current values are both NULL for the ETSEC, MEVAL, and AGEO questions and the section is marked as changed, we will no longer be providing value codes for those disabled questions. Instead, we will leave them blank.

mlan225:
Fixes: #717

Revert the recent changes to the A3 export.

Previously, for the nullable properties of each section: 
- ETSEC
- MEVAL
- AGEO

If the section had been marked as changed, then these properties would need to return a value code in the export if both previous and current values are NULL.

Now, we are reverting the logic to be as follows:
> If a section is marked as changed and the previous value and current value are null for ETSEC, MEVAL, and AGEO, then we should return NULL as the export value


mlan225:
### Review
- [x] Run local tests to check or success
- [x] Speak with Josh to review A3 rules and confirm functionality
- [x] Test A3 export with different data 
- [x] Review relevant code

ashleybot (approved):
@mlan225 Good to bump to next stable patch and merge to main.

PR #728, opened 2026-08-11
Add D1c fields

### LINKED ISSUE #727: Add an export placeholder for new form D1c
If the visit date is on or after July 14, 2026, include these columns with default values in the export:
`d1cnot=‘88’` and `moded1c=‘0’`

We do not yet have a D1c form and are awaiting more information from the Clinical Core before implementing.

CallowaySimon:
Resolves #727

### DIFF
--- a/src/UDS.Net.Forms/Pages/PacketSubmissions/Export.cshtml.cs
+++ b/src/UDS.Net.Forms/Pages/PacketSubmissions/Export.cshtml.cs
@@ -23,6 +23,10 @@ namespace UDS.Net.Forms.Pages.PacketSubmissions
         private readonly IVisitService _visitService;
 
         public bool Processed { get; set; } = false;
+        private static readonly DateTime D1cEffectiveDate = new(2026, 7, 14);
+        private const string D1cNotDefaultValue = "88";
+        private const string ModeD1cDefaultValue = "0";
+
 
         public ExportModel(IPacketService packetService, IParticipationService participationService, IConfiguration configuration, IVisitService visitService)
         {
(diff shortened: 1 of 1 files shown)


src/UDS.Net.Forms/Pages/PacketSubmissions/Export.cshtml.cs
```diff
@@ -47,96 +51,116 @@ public async Task<IActionResult> OnGetAsync(int packetId)
...
-                bool headerWritten = false;
+                var packet = await _packetService.GetPacketWithForms(User.Identity.Name, id);
+
+                if (packet == null)
+                    continue;
+
+                var participant = await _participationService.GetById(User.Identity.Name, packet.ParticipationId);
 
-                foreach (var id in packetId)
+                if (participant == null)
```
ashleybot: @CallowaySimon Let's keep the coding convention to wrap the code in `if (packet != null)` rather than using `continue;`

src/UDS.Net.Forms/Pages/PacketSubmissions/Export.cshtml.cs
```diff
@@ -47,96 +51,116 @@ public async Task<IActionResult> OnGetAsync(int packetId)
...
             memoryStream.Position = 0;
 
             string filename = packetSubmission.ToVM().GetFileName(participant.LegacyId, packet.VISIT_DATE);
 
             Response.Headers["Content-Disposition"] = $"attachment; {filename}";
+
             return File(memoryStream, "text/csv", filename);
         }
 
         public async Task<IActionResult> OnPostExportMultiplePackets(List<int> packetId)
```
ashleybot: @CallowaySimon Let's improve this parameter name to better reflect that it is a list by renaming it to `packetIds`

src/UDS.Net.Forms/Pages/PacketSubmissions/Export.cshtml.cs
CallowaySimon: I have changed the code to follow the if (packet != null) convention, rather than using continue. I have also added a check to make sure that if not all packets are correct, then the whole operation fails prior to packet submission creation. 

The lines 
`csv.WriteField(string.Empty);`
`csv.WriteField(string.Empty);`
are only executed if 
`if (includeD1cColumns)`
So there would be no addition columns if just a single packet pr July 14th was exported.

src/UDS.Net.Forms/Pages/PacketSubmissions/Export.cshtml.cs
CallowaySimon: I have update the parameter to packetIds, and also updated the JS due to the change.

ashleybot (approved):
@CallowaySimon Thanks for making those changes. Good to bump to stable version and merge to main!

